PR #174, opened 2022-09-21
adds serialization/deserialization of private key

Dev43:
Adds serialization and deserialization of the private key. Allows me to save the private key somewhere, to then recreate my public/private key pair.

### DIFF
--- a/seal_fhe/src/key_generator.rs
+++ b/seal_fhe/src/key_generator.rs
@@ -330,11 +330,19 @@ impl SecretKey {
     pub fn get_handle(&self) -> *mut c_void {
         self.handle
     }
+}
 
+impl PartialEq for SecretKey {
+    fn eq(&self, other: &Self) -> bool {
+        self.as_bytes() == other.as_bytes()
+    }
+}
+
+impl ToBytes for SecretKey {
     /**
      * Returns the key as a byte array.
      */
-    pub fn as_bytes(&self) -> Result<Vec<u8>> {
+    fn as_bytes(&self) -> Result<Vec<u8>> {
         let mut num_bytes: i64 = 0;
 
         convert_seal_error(unsafe {
@@ -362,12 +370,6 @@ impl SecretKey {
     }
 }
 
-impl PartialEq for SecretKey {
-    fn eq(&self, other: &Self) -> bool {
-        self.as_bytes() == other.as_bytes()
-    }
-}
-
 impl FromBytes for SecretKey {
     fn from_bytes(context: &Context, bytes: &[u8]) -> Result<Self> {
         let key = SecretKey::new()?;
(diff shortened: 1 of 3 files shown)


seal_fhe/src/key_generator.rs
```diff
@@ -354,6 +362,31 @@ impl SecretKey {
     }
 }
 
+impl PartialEq for SecretKey {
+    fn eq(&self, other: &Self) -> bool {
+        self.as_bytes() == other.as_bytes()
```
rickwebiii, line 367: Thanks for taking this on! I guess I overlooked this while making things serializable.

Move the `as_bytes()` implementation out from `impl SecretKey` to under `impl ToBytes for SecretKey`. This makes it compatible with `WithContext`.

`WithContext` is our wrapper to make artifacts compatible with Serde Serialize/Deserialize traits. It manages first deserializing the SealContext so it can be passed to `from_bytes` when deserializing the data payload.

sunscreen_runtime/src/keys.rs
```diff
@@ -49,6 +49,7 @@ pub struct PublicKey {
 /**
  * The private key used to decrypt ciphertexts.
  */
+#[derive(Serialize)]
```
rickwebiii, line 52: `#[derive(Serialize, Deserialize, Clone, PartialEq)]`.

sunscreen_runtime/src/keys.rs
```diff
@@ -49,6 +49,7 @@ pub struct PublicKey {
...
 pub struct PrivateKey(pub(crate) SealSecretKey);
```
rickwebiii, line 53: This will become
```
pub struct PrivateKey(pub(crate) WithContext<SealSecretKey>);
```

sunscreen_runtime/src/runtime.rs
```diff
@@ -158,6 +158,22 @@ impl Runtime {
         Ok(keys)
     }
 
+    /**
+     * Takes a byte array of the serialized private key material and returns a PrivateKey instance
+     */
+    pub fn bytes_to_private_key(&self, sk: &[u8]) -> Result<PrivateKey> {
```
rickwebiii, line 164: We shouldn't have a special function for deserialization, but should use serde's `Deserialize` trait as all the other keys to. `WithContext` is the key to bridging SEAL's serialization with serde.

rickwebiii (commented):
In runtime.rs, you'll need to update `generate_keys()` to make the `WithContext` type for `SecretKey`. Just follow the same pattern we use for all the other keys.

Additionally, update the test `can_roundtrip_all_keys` to serialize and deserialize the `SecretKey` as well.

Dev43:
Thanks for the feedback, updated it as needed

rickwebiii:
Perfect, check in when the builds pass!

PR #177, opened 2022-09-27
Add chaining docs

rickwebiii:


sunscreen_docs/src/from_programs_to_apps/chaining.md
```diff
@@ -0,0 +1,30 @@
+# Chaining
+
+Suppose you want to run one program and its output as the input to another program. Sunscreen refers to this pattern as *chaining*. For example, suppose we have an account balance and periodically need to add incoming transactions. We can write a simple FHE program that returns the new balance after adding a transaction:
```
rickwebiii, line 3: ... and *use* its output as the input.

rickwebiii, line 3: "We can write a simple FHE program that returns the new balance after adding a transaction"

"We can express this with the following FHE program:"

sunscreen_docs/src/from_programs_to_apps/chaining.md
```diff
@@ -0,0 +1,30 @@
...
+```rust
+#use sunscreen::{fhe_program, types::{Cipher, bfv::Signed};
+
+#[fhe_program(scheme = "bfv", chain_count = 10)]
+fn accumulate_balance(balance: Cipher<Signed>, tx_amount: Cipher<Signed>) -> Cipher<Signed> {
+    balance + tx_amount
+}
+```
+
+The `chain_count = 10` tells the compiler to choose FHE scheme parameters that allow us to chain this program up to 10 times. That is, we can use `accumulate_balance`'s returned ciphertext as the input to another run of the `accumulate_balance` and can repeat this process up to 10 times. After the 10th run, the resulting ciphertext can still be decrypted.
```
rickwebiii, line 25: After the 10th run...

After the final run...

PR #182, opened 2022-10-31
Run clippy in CI

samtay:
This PR
- Runs clippy against all targets in CI, failing the build on any warnings
- Fixes the current warnings.

Most of the changes are small and uncontroversial. However there were a few warnings where I felt the default clippy behavior would hinder readability, or simply weren't relevant for us. In those cases I added ad-hoc `#[allow]` pragmas.

Definitely let me know if you'd like fewer or more things to be `allow`ed. In fact, it may make sense to add a `clippy.toml` for certain things across the board. For example, some of the example programs that simulate lines of equations use large input / output tuples, and clippy complains about type complexity, but I think type synonyms would probably only add confusion there :shrug:  Let me know what you think.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -40,6 +40,9 @@ jobs:
     # Check the submitted change meets style guidelines
     - name: Cargo Format
       run: cargo fmt --check
+    # Check the submitted change passes the clippy linter
+    - name: Cargo Clippy
+      run: cargo clippy --release --all-targets -- --deny warnings
     # Build and run the tests
     - name: Build and run tests
       run: cargo test --verbose --release
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -40,9 +40,6 @@ jobs:
     # Check the submitted change meets style guidelines
     - name: Cargo Format
       run: cargo fmt --check
-    # Check the submitted change passes the clippy linter
-    - name: Cargo Clippy
-      run: cargo clippy --release --all-targets -- --deny warnings
     # Build and run the tests
     - name: Build and run tests
       run: cargo test --verbose --release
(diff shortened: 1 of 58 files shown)


examples/calculator_fractional/src/main.rs
```diff
@@ -44,7 +44,7 @@ struct Expression {
 enum ParseResult {
     Help,
     Exit,
-    Expression(Expression),
+    Expression(Box<Expression>),
```
rickwebiii, line 47: It's pretty cool that Clippy caught this, but for examples we need to value readability over all else. Box is a Rustism we don't want to bog users down with while getting ramped up. Add an ignore for this one.

examples/calculator_rational/src/main.rs
```diff
@@ -43,7 +43,7 @@ struct Expression {
 enum ParseResult {
     Help,
     Exit,
-    Expression(Expression),
+    Expression(Box<Expression>),
```
rickwebiii, line 46: See above.

examples/chi_sq/src/main.rs
```diff
@@ -11,6 +11,8 @@
 //! Sunscreen's results with other FHE compilers, see
 //! [SoK: Fully Homomorphic Encryption Compilers](https://arxiv.org/abs/2101.07078).
 
+#![allow(clippy::type_complexity)]
```
rickwebiii, line 14: These are scope-able, right? Clippy is complaining that `Cipher<Signed>` is a very complex type definition. I wonder if this is due to Cippy evaluating the expanded type that appears after the `#[fhe_program]` gets expanded, which is `CircuitNode<Cipher<Signed>>`.

If so, we should probably add this allow to `fhe_program_impl` in `sunscreen_comiler_macros/src/fhe_program.rs`. This way, users won't get this clippy warning on our account.

seal_fhe/src/modulus.rs
```diff
@@ -71,9 +71,9 @@ impl TryFrom<i32> for SecurityLevel {
     }
 }
 
-impl Into<i32> for SecurityLevel {
-    fn into(self) -> i32 {
-        match self {
+impl From<SecurityLevel> for i32 {
```
rickwebiii, line 74: TIL, this doesn't violate the foreign traits rules.

sunscreen/src/types/bfv/batched.rs
```diff
@@ -237,7 +237,7 @@ impl<const LANES: usize> TryFromPlaintext for Batched<LANES> {
                 row_0
                     .iter()
                     .take(LANES)
-                    .map(|x| *x)
+                    .copied()
```
rickwebiii, line 240: TIL

.github/workflows/rust.yml
```diff
@@ -40,6 +40,9 @@ jobs:
     # Check the submitted change meets style guidelines
     - name: Cargo Format
       run: cargo fmt --check
+    # Check the submitted change passes the clippy linter
```
rickwebiii, line 43: It might be worth making a new job for this, as it adds like 30s to our build. Then it will run in parallel on another build agent.

rickwebiii (approved):
Looks good overall. A few comments about the examples.

PR #185, opened 2022-11-15
Rweber/zkp

rickwebiii:
Sorry for the mega diff :( I'm checking this in now to prevent it from growing even larger.

This refactoring does the following:
* Updates the FHE compiler to use the shared Context and FrontendCompilation types in sunscreen_compiler_common.
* Adds the start of the ZKP frontend compiler that also uses these types
* Adds basic support for ZKP NativeFields


sunscreen_bulletproofs_backend/src/lib.rs
```diff
@@ -0,0 +1,77 @@
+use bulletproofs::{
```
rickwebiii, line 1: Idea: rather than have N crates for each backend, have a single ZKP backend crate with each supplied backend being a feature. Expose traits and types needed to implement a backend in this crate so others can make their own backends.

rickwebiii: Also, this crate is currently nonsense and isn't hooked up to the frontend.

sunscreen_compiler_common/src/context.rs
```diff
@@ -98,6 +101,20 @@ where
     }
 }
 
+impl<O> PartialEq for FrontendCompilation<O>
```
rickwebiii, line 104: This is really for testing, as graph isomorphism is generally an NP-complete problem. Add a comment as such.

petgraph uses the VF2 algortithm, which is minimally O(V^2) and O(V!*V) in the worst case (lol).

samtay (approved):

PR #186, opened 2022-11-18
Rweber/zkp

rickwebiii:
Add Bulletproofs backend for ZKP compiler

### DIFF
--- a/sunscreen_compiler_common/src/graph.rs
+++ b/sunscreen_compiler_common/src/graph.rs
@@ -19,7 +19,7 @@ pub struct GraphQuery<'a, N, E>(&'a StableGraph<N, E>);
 
 impl<'a, N, E> From<&'a StableGraph<N, E>> for GraphQuery<'a, N, E> {
     fn from(x: &'a StableGraph<N, E>) -> Self {
-        Self(&x)
+        Self(x)
     }
 }
 
@@ -382,7 +382,7 @@ where
             return Err(GraphQueryError::IncorrectBinaryOperandEdges);
         }
 
-        let left = parent_edges.iter().next();
+        let left = parent_edges.first();
 
         Ok(left
             .ok_or(GraphQueryError::IncorrectUnaryOperandEdge)?
--- a/sunscreen_compiler_common/src/graph.rs
+++ b/sunscreen_compiler_common/src/graph.rs
@@ -71,6 +71,12 @@ where
     fn apply(&mut self, graph: &mut StableGraph<N, E>);
 }
 
+// Make a surrogate implementation of the trait for traversal functions
+// that don't mutate the graph.
+impl<N, E> TransformList<N, E> for () where N: Clone, E: Clone {
+    fn apply(&mut self, graph: &mut StableGraph<N, E>) { }
+}
+
 /**
  * A specialized topological DAG traversal that allows the following graph
  * mutations during traversal:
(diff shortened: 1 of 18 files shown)


sunscreen_zkp_backend/src/bulletproofs.rs
```diff
@@ -1,77 +1,375 @@
...
+            &[(in_2, EdgeInfo::Left), (mul_1, EdgeInfo::Right)],
+        );
+
+        let _ = add_node(
+            BackendOperation::Constraint(BigInt::from_u32(42)),
+            &[(add_1, EdgeInfo::Unordered)],
+        );
+
+        // 10 * 4 + 2 == 42
+        let proof = BulletproofsR1CSCircuit::prove(
```
rickwebiii, line 363: Add another pass case and a fail case.

sunscreen_zkp_backend/src/bulletproofs.rs
```diff
@@ -1,77 +1,375 @@
...
+
+        let mut verifier = Verifier::new(transcript);
+
+        circuit.gen_circuit(&mut graph, &mut verifier, |_| None)?;
+
+        Ok(verifier.verify(&proof.0, &pedersen_gens, &bulletproof_gens)?)
+    }
+}
+
+fn try_uint_to_scalar<const N: usize>(x: &UInt<N>) -> Result<Scalar> {
```
rickwebiii, line 223: I guess I should see if the foreign trait rules allow us to `impl TryFrom<>`

sunscreen_zkp_backend/src/bulletproofs.rs
```diff
@@ -1,77 +1,375 @@
...
+        })?;
+
+        Ok(())
     }
+}
 
-    pub fn verify(&self) -> bool {
-        let (transcript, pc_gens, bp_gens) = Self::make_gens();
+impl ZkpProverBackend for BulletproofsR1CSCircuit {
+    fn prove(mut graph: ZkpBackendCompilationResult, inputs: &[BigInt]) -> Result<Box<Proof>> {
```
rickwebiii, line 155: Kinda gross that graph has to be passed as mut when in fact it won't be mutated.

This is because I didn't want to implement an immutable traversal function, so we're using the mutable one and just not returning any transforms so the graph doesn't mutate.

sunscreen_zkp_backend/src/bulletproofs.rs
```diff
@@ -1,77 +1,375 @@
...
 
-    fn gadget<CS: ConstraintSystem>(cs: &mut CS, inputs: Vec<Variable>, outputs: Vec<Variable>) {
-        let (_, _, o) = cs.multiply(
-            inputs[0] + Scalar::from(0u32),
-            inputs[1] + Scalar::from(0u32),
-        );
+impl ZkpVerifierBackend for BulletproofsR1CSCircuit {
+    fn verify(mut graph: ZkpBackendCompilationResult, proof: Box<Proof>) -> Result<()> {
+        dbg!(proof.type_id());
+        dbg!(TypeId::of::<BulletproofsR1CSProof>());
```
samtay, line 195: A few leftover `dbg`s

sunscreen_zkp_backend/src/bulletproofs.rs
rickwebiii: Fixed.

sunscreen_compiler_common/src/graph.rs
```diff
@@ -218,3 +251,177 @@ where
...
+     * Returns the unordered operands to the given operation.
+     *
+     * # Remarks
+     * As these operands are unordered, their order is undefined. Use
+     * [`EdgeInfo::Ordered`] if you need a defined order.
+     *
+     * * # Errors
+     * - No node exists at the given index.
+     * - The node at the given index isn't a unary operation.
+     * - The node at the given index doesn't have a single unary operand.
```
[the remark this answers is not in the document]
rickwebiii: That's what it *should* do. I've updated the function to check that the single operand is unary.

sunscreen_compiler_common/src/graph.rs
```diff
@@ -218,3 +251,177 @@ where
...
+
+        let parent_edges = self
+            .edges_directed(index, Direction::Incoming)
+            .collect::<Vec<EdgeReference<EdgeInfo>>>();
+
+        if !parent_edges
+            .iter()
+            .all(|e| matches!(e.weight(), EdgeInfo::Unordered))
+        {
+            return Err(GraphQueryError::IncorrectUnaryOperandEdge);
```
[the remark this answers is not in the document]
rickwebiii: You are correct. Bad copypasta.

sunscreen_compiler_common/src/transforms/common_subexpression_elimination.rs
```diff
@@ -131,13 +132,14 @@ pub fn common_subexpression_elimination<O: Operation>(
             }
         }
 
-        transforms
-    });
+        Ok(transforms) as Result<GraphTransforms<NodeInfo<O>, EdgeInfo>, Infallible>
```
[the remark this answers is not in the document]
rickwebiii: Changed. I'm still not 100% on when you can and can't use _ in type matching, but `Ok::<_, Infallable>()` is definitely cleaner.

sunscreen_zkp_backend/src/bulletproofs.rs
rickwebiii: You can't.

PR #187, opened 2022-11-22
Rweber/zkp

rickwebiii:
Refactor API to facilitate ZKPs

### DIFF
--- a/examples/amm/src/main.rs
+++ b/examples/amm/src/main.rs
@@ -1,7 +1,7 @@
 use sunscreen::{
     fhe_program,
     types::{bfv::Rational, Cipher},
-    Ciphertext, CompiledFheProgram, Compiler, Error, FheRuntime, GenericRuntime, Params,
+    Ciphertext, CompiledFheProgram, Compiler, Error, FheRuntime, Runtime, Params,
     PrivateKey, PublicKey,
 };
 
@@ -28,7 +28,7 @@ impl Miner {
     pub fn setup() -> Result<Miner, Error> {
         let app = Compiler::new().fhe_program(swap_nu).compile()?;
 
-        let runtime = GenericRuntime::new_fhe(app.params())?;
+        let runtime = Runtime::new_fhe(app.params())?;
 
         Ok(Miner {
             compiled_swap_nu: app.get_fhe_program(swap_nu).unwrap().clone(),
@@ -63,7 +63,7 @@ struct Alice {
 
 impl Alice {
     pub fn setup(params: &Params) -> Result<Alice, Error> {
-        let runtime = GenericRuntime::new_fhe(params)?;
+        let runtime = Runtime::new_fhe(params)?;
 
         let (public_key, private_key) = runtime.generate_keys()?;
 
--- a/examples/amm/src/main.rs
+++ b/examples/amm/src/main.rs
@@ -1,8 +1,8 @@
 use sunscreen::{
     fhe_program,
     types::{bfv::Rational, Cipher},
-    Ciphertext, CompiledFheProgram, Compiler, Error, FheRuntime, Runtime, Params,
-    PrivateKey, PublicKey,
+    Ciphertext, CompiledFheProgram, Compiler, Error, FheRuntime, Params, PrivateKey, PublicKey,
+    Runtime,
 };
 
 #[fhe_program(scheme = "bfv")]
--- a/examples/amm/src/main.rs
+++ b/examples/amm/src/main.rs
@@ -1,7 +1,8 @@
 use sunscreen::{
     fhe_program,
     types::{bfv::Rational, Cipher},
-    Ciphertext, CompiledFheProgram, Compiler, Error, Params, PrivateKey, PublicKey, Runtime,
+    Ciphertext, CompiledFheProgram, Compiler, Error, FheRuntime, GenericRuntime, Params,
+    PrivateKey, PublicKey,
 };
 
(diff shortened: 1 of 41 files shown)


sunscreen/src/fhe_compiler.rs
```diff
@@ -46,117 +47,85 @@ pub trait FheProgramFn {
...
         Self::new()
     }
 }
 
-impl Compiler {
+/**
+ * A private method for converting between Compiler types. We don't want
+ * to expose this as part of the public API, hence we don't `impl From`.
+ */
+fn into<T, U>(x: Compiler<T>) -> Compiler<U> {
```
samtay, line 70: Does it make sense to have this exist for all `T, U`? It's internal so it's not a huge deal, but if only certain conversions make sense, we could restrict this at the type level.

sunscreen_docs/src/advanced/batching/dot_product.md
```diff
@@ -250,7 +250,7 @@ fn main() {
 
     // 3. Make our runtime, generate keys, encrypt our data, run our
     // program, and check the result
-    let runtime = Runtime::new(&fhe_program.metadata.params).unwrap();
+    let runtime = Runtime::new_fhe(&fhe_program.metadata.params).unwrap();
```
rickwebiii, line 253: Got through each md file and make sure the reading is still consistent with the examples.

sunscreen/src/lib.rs
```diff
@@ -105,39 +110,104 @@ impl Application {
...
+        Self {
+            fhe_programs: other.fhe_programs,
+            zkp_programs: other.zkp_programs,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl Application<FheZkp> {
+    pub(crate) fn to_application(other: Application<()>) -> Application<FheZkp> {
```
samtay, line 205: These impls seem odd. It seems like anytime someone uses `Application::new()` they get back an `app: Application<()>` and then need to use `Application<Something>::to_application(app)`? I imagine we can simplify this. 

In general I wonder if these `Struct<()>` impls with `new` and whatnot could instead just be generic. Then someone can specify what they want by just specifying the generic, e.g. `Application<Fhe>::new().get_fhe_program()`.

sunscreen_runtime/src/runtime.rs
```diff
@@ -20,21 +22,77 @@ enum Context {
...
+    fn unwrap_fhe(&self) -> &FheRuntimeData {
+        match self {
+            Self::Fhe(x) => x,
+            _ => panic!("Expected RuntimeData::Fhe."),
+        }
+    }
+}
+
+/**
+ * Contains all the elements needed to encrypt, decrypt, generate keys, and evaluate FHE programs.
```
samtay, line 85: This comment indicates the type is specific to FHE programs, but the type looks more general than that.

examples/amm/src/main.rs
```diff
@@ -20,17 +21,17 @@ struct Miner {
...
-    runtime: Runtime,
+    runtime: FheRuntime,
 }
 
 impl Miner {
     pub fn setup() -> Result<Miner, Error> {
         let app = Compiler::new().fhe_program(swap_nu).compile()?;
 
-        let runtime = Runtime::new(app.params())?;
+        let runtime = GenericRuntime::new_fhe(app.params())?;
```
samtay, line 31: All of these examples import `GenericRuntime`, but the docs in that module all reference the type synonym `Runtime`. I think the examples and docs should be in agreement on the intended use. (Personally I'd keep it as `Runtime` throughout.)

sunscreen/src/fhe_compiler.rs
rickwebiii: The cases this needs to handle are:
`Compiler<()> -> Compiler<Zkp>`
`Compiler<()> -> Compiler<Fhe>`
`Compiler<Zkp> -> Compiler<FheZkp>`
`Compiler<Fhe> -> Compiler<FheZkp>`

To add bounds to make all of these work, I think we would need to add a new trait and not use (), but some other marker type. I don't think it's worth it.

This function is indeed internal only.

sunscreen/src/lib.rs
rickwebiii: This should be an `into()` function as we did with Compiler. But the idea is the same: our code internally needs to convert between types when appropriate, but users never do.

Users never construct Applications themselves (the `new` method is marked pub(crate)). Rather, the compiler emits the appropriately typed one.

samtay:
Also, reminder to update the playground as part of this PR. Here's a [branch](https://github.com/Sunscreen-tech/rust-playground/compare/sunscreen...Sunscreen-tech:rust-playground:samtay/rw-zkp) with the current state of the `pir` example if you wanna just use that commit.

examples/amm/src/main.rs
rickwebiii: This was a bad F2 refactor. I've fixed this.

sunscreen_docs/src/advanced/batching/dot_product.md
rickwebiii: As discussed in our sync, we'll do this after checking in.

sunscreen/src/lib.rs
rickwebiii: I've migrated to the same pattern as into with a method on

```
impl<T> Application<T> {
  fn convert<U>(self) -> Application<U> { ... }
}
```

PR #189, opened 2022-11-29
Rweber/clippy

rickwebiii:
Reduce size of errors and fix clippy issues. Our largest errors are now 24 bytes, down from 144.

### DIFF
--- a/sunscreen_fhe_program/src/error.rs
+++ b/sunscreen_fhe_program/src/error.rs
@@ -1,4 +1,5 @@
 use petgraph::stable_graph::NodeIndex;
+use static_assertions::const_assert;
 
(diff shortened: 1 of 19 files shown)


sunscreen_fhe_program/src/error.rs
```diff
@@ -60,15 +104,64 @@ pub enum NodeError {
...
 
     /**
      * The node has expects a specific number of input operands (first argument),
      * but got some other number (second argument).
      */
-    WrongOperandCount(usize, usize),
+    WrongOperandCount(Box<(usize, usize)>),
+}
+
+impl std::fmt::Display for NodeError {
```
rickwebiii, line 116: Can use the unwrap technique in RuntimeError

PR #196, opened 2022-12-14
Rweber/zkp

rickwebiii:
WIP: Add polynomial type for use in ZKP programs.

### DIFF
--- a/sunscreen/src/types/zkp/rns_polynomial.rs
+++ b/sunscreen/src/types/zkp/rns_polynomial.rs
@@ -1,0 +1,149 @@
+use sunscreen_compiler_macros::TypeName;
+use sunscreen_runtime::ZkpProgramInputTrait;
+use sunscreen_zkp_backend::BigInt;
+
+use crate::{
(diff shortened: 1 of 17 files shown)


sunscreen/src/types/zkp/rns_polynomial.rs
```diff
@@ -0,0 +1,156 @@
...
+                    .map(|x| {
+                        x.into_iter()
+                            .map(|x| x.into())
+                            .collect::<Vec<NativeField>>()
+                            .try_into()
+                            .unwrap()
+                    })
+                    .collect::<Vec<[NativeField; N]>>()
+                    .try_into()
+                    .unwrap(),
```
samtay, line 46: Is this code also temporary? If not, might be better to impl TryFrom/TryInto and unwrap at call site, unless we know this is infallible.

rickwebiii: TIL, `array.map([T; N]) -> [U; N]`. That cleans the iteration nonsense up.

PR #197, opened 2022-12-22
Rweber/zkp

rickwebiii:
Move `BackendField` into type system so we have access to the field modulus.

### DIFF
--- a/sunscreen/src/compiler.rs
+++ b/sunscreen/src/compiler.rs
@@ -383,7 +383,7 @@ impl FheCompiler {
      * will return a [`Error::NameCollision`] error.
      */
     pub fn compile(self) -> Result<Application<Fhe>> {
-        Ok(Application::new(self.compile_fhe()?, HashMap::new())?)
+        Application::new(self.compile_fhe()?, HashMap::new())
     }
 }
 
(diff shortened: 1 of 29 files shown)


sunscreen/tests/simd.rs
```diff
@@ -275,7 +275,7 @@ fn can_neg_cipher_cipher() {
         neg_impl(a)
     }
 
-    let app = Compiler::new()
+    let app = GenericCompiler::new()
```
rickwebiii, line 278: Just use `Compiler`.

sunscreen/src/compiler.rs
```diff
@@ -8,7 +8,7 @@ use std::collections::{HashMap, HashSet};
 use std::marker::PhantomData;
 use sunscreen_fhe_program::FheProgramTrait;
 use sunscreen_runtime::{marker, CompiledFheProgram, Fhe, FheZkp, Zkp};
-use sunscreen_zkp_backend::CompiledZkpProgram;
+use sunscreen_zkp_backend::{BackendField, CompiledZkpProgram, ZkpBackend};
```
rickwebiii, line 11: TODO: Prevent user from calling compile until they have specified either a ZKP or FHE program.

Doing so today returns `Error::NoPrograms` at runtime, but we can make this a compile-time error.

rickwebiii: Another time.

PR #198, opened 2023-01-05
Can encode negative numbers onto native fields

rickwebiii:


sunscreen/src/types/zkp/native_field.rs
```diff
@@ -49,18 +49,78 @@ impl<F: BackendField> NativeField<F> {
...
+    fn from(x: u16) -> Self {
+        (x as u64).into()
+        
+    }
+}
+
+impl<F: BackendField> From<u32> for NativeField<F>
+{
+    fn from(x: u32) -> Self {
+        (x as u64).into()
```
samtay, line 71: Not a huge deal, but these `x as u64`s rely on the programmer verifying that `x` fits into `u64`. I think for this `From` instance `x.into()` is more appropriate, letting the compiler verify this instead.

rickwebiii: Fixed.

PR #204, opened 2023-01-18
Rweber/zkp

rickwebiii:


sunscreen/src/types/zkp/gadgets/arithmetic.rs
```diff
@@ -0,0 +1,221 @@
...
+    ) -> ZkpResult<Vec<sunscreen_zkp_backend::BigInt>> {
+        let x = gadget_inputs[0];
+        let m = gadget_inputs[1];
+
+        // As per docs, when shift amount is constant, time is constant.
+        // See https://docs.rs/crypto-bigint/latest/crypto_bigint/struct.UInt.html#method.shr_vartime
+        let fm_2 = self.field_modulus.shr_vartime(2);
+
+        let is_neg = x.ct_gt(&fm_2);
+        let (q_pos, r_pos) = x.div_rem(&m).unwrap();
```
samtay, line 92: So this will panic iff `m == 0`. I don't see us asserting that is the case anywhere, though. Should we add that? Ditto on the `reduce` unwraps below.

If this isn't assertable anywhere, would it make more sense to have this bubble up to a ZkpResult instead of panicking?

rickwebiii: I actually did have a panic in previous iterations where m was passed in the gadget constructor. I then changed it to just receive `m` as an argument and forgot to add the error condition.

rickwebiii: Good catch either way.

PR #205, opened 2023-01-20
Rweber/zkp

rickwebiii:
Can prove BFV encryptions in R1CS.

### DIFF
--- a/sunscreen/src/types/zkp/gadgets/binary.rs
+++ b/sunscreen/src/types/zkp/gadgets/binary.rs
@@ -10,7 +10,17 @@ pub struct ToUInt {
 }
 
 impl ToUInt {
+    /**
+     * Creates a new [`ToUInt`] gadget.
+     * 
+     * # Panics
+     * * If n > 512
+     */
     pub fn new(n: usize) -> Self {
+        if n > 512 {
+            panic!("Cannot decompose into > 512 bit values.");
+        }
+
         Self { n }
     }
 }
@@ -21,6 +31,10 @@ impl Gadget for ToUInt {
 
         if self.n == 0 {
             return Err(ZkpError::gadget_error("Cannot create 0-bit uint."));
+        }
+
+        if self.n > 512 {
+            return Err(ZkpError::gadget_error("Cannot decompose into > 512-bit values."));
         }
 
         if *val > BigInt::ONE.shl_vartime(self.n) {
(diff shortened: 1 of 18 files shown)


sunscreen/src/types/zkp/gadgets/binary.rs
```diff
@@ -23,6 +33,12 @@ impl Gadget for ToUInt {
             return Err(ZkpError::gadget_error("Cannot create 0-bit uint."));
         }
 
+        if self.n > 512 {
+            return Err(ZkpError::gadget_error(
```
samtay, line 37: These seems redundant since it is checked in the constructor, but I suppose it doesn't hurt

sunscreen/src/types/zkp/rns_polynomial.rs
```diff
@@ -134,15 +142,61 @@ impl<F: BackendField, const N: usize, const R: usize> MulVar for RnsRingPolynomi
...
+
+impl<F: BackendField, const D: usize, const R: usize> Mod<F> for RnsRingPolynomial<F, D, R> {
+    fn signed_reduce(
+        lhs: ProgramNode<Self>,
+        m: ProgramNode<NativeField<F>>,
+        remainder_bits: usize,
+    ) -> ProgramNode<Self> {
+        let residues = lhs.residues();
+
+        let mut outputs = vec![];
```
samtay, line 179: optionally `Vec::with_capacity(D)` ; though I'm not sure how important it is to micro-optimize the compilation step.

rickwebiii: I did some basic profiling and found a way bigger problem than this; we have O(N^2) behavior when traversing the build graph.

sunscreen/src/types/zkp/rns_polynomial.rs
```diff
@@ -254,4 +308,58 @@ mod tests {
 
         assert!(result.is_err());
     }
+
+    #[test]
+    fn can_scale_polynomial() {
+        #[zkp_program(backend = "bulletproofs")]
+        fn add_poly<F: BackendField>(
```
samtay, line 315: I think this name was copy-pasted, as the function itself doesn't do any polynomial addition.

rickwebiii: Yep

sunscreen/src/types/zkp/gadgets/binary.rs
rickwebiii: Removed

PR #207, opened 2023-01-23
Fix O(N^2) behavior in graph traversal

rickwebiii:


sunscreen_compiler_common/src/transforms/graph_transforms.rs
```diff
@@ -161,8 +168,14 @@ where
                 }
             };
 
+            if let Some(node) = inserted_node {
```
rickwebiii, line 171: Now that I think about it, the caller can just iterate through `inserted_node_ids` and look for `Some(_)` entries.

samtay (approved):

sunscreen_zkp_backend/src/bulletproofs.rs
```diff
@@ -170,56 +186,66 @@ impl BulletproofsCircuit {
...
-                        .unwrap_or_else(|| panic!("{}", dependency_not_found_msg(left)))
+                        .unwrap_or_else(|| panic!("{}", dependency_not_found_msg(left_idx)))
                         .clone();
 
-                    let right = self.nodes[right.index()]
+                    let right = self.nodes[right_idx.index()]
                         .as_ref()
-                        .unwrap_or_else(|| panic!("{}", dependency_not_found_msg(right)))
+                        .unwrap_or_else(|| panic!("{}", dependency_not_found_msg(right_idx)))
                         .clone();
```
samtay, line 202: so is that vec internal to the bulletproof obj? i'm wondering if we can just put an Rc around it and get this behavior for free. that way even these clones are just bumping reference counts:
```
                    let right = self.nodes[right_idx.index()]
                        .as_ref()
                        .unwrap_or_else(|| panic!("{}", dependency_not_found_msg(right_idx)))
                        .clone();
```

sunscreen_zkp_backend/src/bulletproofs.rs
```diff
@@ -145,10 +145,26 @@ impl BulletproofsCircuit {
...
         // The graph won't actually be mutated.
         forward_traverse(&graph.0, |query, idx| {
             let node = query.get_node(idx).unwrap();
 
+            // Each linear combination object in Bulletproofs has a Vec
+            // in it and thus ain't cheap to store. As such, we reference
+            // count the output of a given node when all its children have
+            // been processed.
+            let ref_count = |nodes: &mut Vec<Option<Node>>, idx: NodeIndex, unprocessed_child_count: &mut Vec<usize>| {
+                unprocessed_child_count[idx.index()] -= 1;
```
samtay, line 164: Is it logically impossible for any of these entries `unprocessed_child_count[idx.index()]` to start at `0`?

samtay: If not, I don't think the logic really changes, just calling it out that we'd want a saturating sub, since these are `usize`s

rickwebiii: The only nodes with an initialized zero count are those with no children. However, this code will never get called in such cases, because there are no children to call it 😄

sunscreen_zkp_backend/src/bulletproofs.rs
rickwebiii: It's not immediately obvious to me how you would do this. What data structure would receive the Rc that doesn't need to be explicitly dropped itself?

samtay: I was thinking something like
```rust
enum Node {
    LinearCombination(Rc<LinearCombination>),
    Scalar(Scalar),
}
```
that way any large variants of the nodes don't blow up as they get cloned. I just played around with this and got it to compile, but I'm not sure it makes any sense. It requires you to `lc.as_ref().clone()` whenever you want to use the `Node::LInearCombination(lc)` in a constraint.

PR #208, opened 2023-01-25
Randomness range proof in R1CS now correct

rickwebiii:
I had to move signed values in F_q to being signed values in F_s, then we can simply add the bound on the gaussian and check the range is [0, 2B). 2B is unfortunately not a power of 2, so the bound is still a little loose, but by less than a factor of 2 (before it was a factor of `q` lol). Could add another `lt` check to make it tight, but I think I'm done messing with this proof for now. We got what we wanted out of it.

### DIFF
--- a/benchmarks/bfv_zkp/src/bfv.rs
+++ b/benchmarks/bfv_zkp/src/bfv.rs
@@ -12,7 +12,7 @@ use sunscreen_zkp_backend::{bulletproofs::BulletproofsBackend, BigInt as ZkpBigI
 
 use crate::poly_ring::PolyRing;
 
-const POLY_DEGREE: usize = 2;
+const POLY_DEGREE: usize = 4;
 
 #[derive(MontConfig)]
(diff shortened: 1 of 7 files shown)


benchmarks/bfv_zkp/src/bfv.rs
```diff
@@ -309,6 +309,60 @@ fn into_rns_poly<F: MontConfig<N>, const N: usize>(
...
+    q_div_2.div2();
+    let q_div_2 = ark_bigint_to_zkp_bigint(q_div_2);
+    let q = ark_bigint_to_zkp_bigint(F::MODULUS);
+
+    assert!(N <= 8);
+    let mut q = ZkpBigInt::ZERO;
+
+    for i in 0..N {
+        q.0.as_words_mut()[i] = F::MODULUS.0[i];
+    }
```
samtay, line 339:
```suggestion
```

samtay (approved):

PR #211, opened 2023-02-10
Rweber/zkp

rickwebiii:
GPU accelerated scalar multiplications! This is the metal implementation...


sunscreen_math/src/metal_impl/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+    include_bytes!(concat!(env!("OUT_DIR"), "/curve25519-dalek.test.metallib"));
+
+pub struct Runtime {
+    device: Device,
+    lib: Library,
+    command_queue: CommandQueue,
+}
+
+/// TODO Remove
+unsafe impl Sync for Runtime {}
```
ryanorendorff, line 34: More detail on why it is here now and what needs to be done to remove it.

sunscreen_math/src/metal_impl/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+        dbg!(gpu_fn.max_total_threads_per_threadgroup());
+
+        encoder.set_compute_pipeline_state(&gpu_fn);
+
+        for (i, buf) in data.iter().enumerate() {
+            encoder.set_buffer(i as u64, Some(buf), 0);
+        }
+
+        let global = MTLSize::new(grid[0].0, grid[1].0, grid[2].0);
+        let local = MTLSize::new(grid[0].1, grid[1].1, grid[2].1);
```
ryanorendorff, line 81: Do we reuse 3x3 grids often? It might be worth making a type alias (or a newtype) for the grid. Plus it is easier for the caller to know what the `[(u64, u64); 3]` represents then.

sunscreen_math/src/metal_impl/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+
+        // TODO: We're relying on non-uniform thread groups. We
```
ryanorendorff, line 83: Question, do we have a style of making an issue and sticking the issue number in the comment? I've seen that procedure before, it can be helpful to keep track of the todos.

sunscreen_math/src/metal_impl/ristrettovec.rs
```diff
@@ -0,0 +1,716 @@
...
+    /// struct EdwardsPoint {
+    ///     X: FieldElement2625,
+    ///     Y: FieldElement2625,
+    ///     Z: FieldElement2625,
+    ///     T: FieldElement2625,
+    /// }
+    ///
+    /// struct FieldElement2625([u32; 10]);
+    /// ```
+    /// To achieve this layout, you must use the u32 backend.
```
ryanorendorff, line 36: Maybe mention that the conversion from u64 to u32 function exists.

sunscreen_math/src/metal_impl/ristrettovec.rs
```diff
@@ -0,0 +1,716 @@
...
+    ///
+    pub fn new(x: &[RistrettoPoint]) -> Self {
+        let runtime = Runtime::get();
+
+        let len = x.len();
+
+        assert_eq!(size_of::<RistrettoPoint>(), size_of::<u32>() * 40);
```
ryanorendorff, line 43: Is this not always true? Do we expect the size of a RistrettoPoint to change upstream? If so, it might be better as a debug assert.

sunscreen_math/src/metal_impl/ristrettovec.rs
```diff
@@ -0,0 +1,716 @@
...
+
+        let mut field_vec = Self { data, len };
+
+        let data_slice = field_vec.buffer_slice_mut();
+
+        for (i, p) in x.iter().enumerate() {
+            let x = p.0.X.to_u29();
+            let y = p.0.Y.to_u29();
+            let z = p.0.Z.to_u29();
+            let t = p.0.T.to_u29();
```
ryanorendorff, line 55: I'm just curious: what is in the other fields of the `RistrettoPoint` (since `.0` was used)?

sunscreen_math/src/metal_impl/ristrettovec.rs
```diff
@@ -0,0 +1,716 @@
...
+            y[i] = buffer_slice[(i + 10) * self.len + index];
+        }
+
+        for i in 0..10 {
+            z[i] = buffer_slice[(i + 20) * self.len + index];
+        }
+
+        for i in 0..10 {
+            t[i] = buffer_slice[(i + 30) * self.len + index];
+        }
```
ryanorendorff, line 111: We seem to be doing the same indexing here as above: is there a way to abstract it out? It might make it easier to reason about the magic numbers like 10 as well.

sunscreen_math/src/metal_impl/ristrettovec.rs
```diff
@@ -0,0 +1,716 @@
...
+
+    pub fn iter(&self) -> RistrettoPoints {
+        RistrettoPoints {
+            vec: self,
+            i: 0usize,
+        }
+    }
+
+    // TODO: This probably needs to return a slice of MaybeUninit<u32>.
+    fn buffer_slice_mut(&mut self) -> &mut [u32] {
```
ryanorendorff, line 129: Do you want to do this now? Or is this comment not true anymore?

sunscreen_math/src/metal_impl/ristrettovec.rs
```diff
@@ -0,0 +1,716 @@
...
+            len: rhs.len(),
+        };
+
+        let len_gpu = U32Arg::new(rhs.len() as u32);
+
+        // TODO: o gets mutated here. Need to figure out what that means in terms of UB.
+        runtime.run(
+            "ristretto_add",
+            &[&self.data, &rhs.data, &o.data, &len_gpu.data],
+            [(rhs.len() as u64, 64), (1, 1), (1, 1)],
```
ryanorendorff, line 187: What does "in terms of UB" mean? Upper bound?

sunscreen_math/src/metal_impl/ristrettovec.rs
```diff
@@ -0,0 +1,716 @@
...
+        // TODO: o gets mutated here. Need to figure out what that means in terms of UB.
+        runtime.run(
+            "ristretto_sub",
+            &[&self.data, &rhs.data, &o.data, &len_gpu.data],
+            [(rhs.len() as u64, 64), (1, 1), (1, 1)],
+        );
+
+        o
+    }
+}
```
ryanorendorff, line 244: Can this at all be simplified by calling `add` but negating the second argument? Does that add overhead?

sunscreen_math/src/metal_impl/ristrettovec.rs
```diff
@@ -0,0 +1,716 @@
...
+
+        runtime.run(
+            "ristretto_scalar_mul",
+            &[&self.data, &rhs.data, &out.data, &len_gpu.data],
+            [(self.len() as u64, 64), (1, 1), (1, 1)],
+        );
+
+        out
+    }
+}
```
ryanorendorff, line 293: The three operations on Ristretto points and scalar vecs seem to follow the same pattern. Potentially this can be abstracted and then called with the individual strings for "ristretto_scalar_mul"/etc.

I could see this being more painful than it is worth. The same idea is repeated several times so it seems ripe for DRY but if we never use the same pattern again then this seems fine.

ryanorendorff (commented):
This looks great!

I am curious whether there if we should do something to make defining `ScalarVec` and `RistrettoPointVec` in a similar way. They are very similar structures with a few differences in `new` and methods like `get`. It would be convenient to write once if possible through something like a trait and then implement the trait for either case. I am not sure how often we will repeat this idiom though (how many `Vec` containers will we need?).

sunscreen_math/src/metal_impl/ristrettovec.rs
```diff
@@ -0,0 +1,716 @@
...
+
+        let buffer = runtime.alloc(self.len_bytes());
+
+        unsafe {
+            std::ptr::copy_nonoverlapping(
+                self.data.contents() as *const u8,
+                buffer.contents() as *mut u8,
+                self.len_bytes(),
+            )
+        };
```
samtay, line 307: Just curious, why do we need this over, say, a default `Clone` impl?

sunscreen_math/src/metal_impl/ristrettovec.rs
```diff
@@ -0,0 +1,716 @@
...
+
+impl<'a> Iterator for RistrettoPoints<'a> {
+    type Item = RistrettoPoint;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let ret = if self.i < self.vec.len() {
+            Some(self.vec.get(self.i))
+        } else {
+            None
+        };
```
samtay, line 332: No need to change this, imo it is clearer as you have written it, but just thought I'd share a common Rust-ism [`bool.then(f)`](https://doc.rust-lang.org/std/primitive.bool.html#method.then)

sunscreen_math/src/metal_impl/ristrettovec.rs
```diff
@@ -0,0 +1,716 @@
...
+        ]);
+
+        let b = ScalarVec::new(&[
+            Scalar::random(&mut thread_rng()),
+            Scalar::random(&mut thread_rng()),
+            Scalar::random(&mut thread_rng()),
+            Scalar::random(&mut thread_rng()),
+        ]);
+
+        /*
```
samtay, line 492: leftover commented code

sunscreen_math/src/metal_impl/ristrettovec.rs
samtay: Curious about this too.. my guess is that the operations are specific to unsigned arithmetic?

sunscreen_math/src/metal_impl/ristrettovec.rs
samtay: Agree, a few well-named variables/consts would probably help this seem less magical/arbitrary.

samtay (changes requested):
Eyes glazed over the `metal` code tbh but hopefully you and @ryanorendorff went over most of that yesterday.

- Some leftover commented code to remove
- CI updates can happen in another PR (I can also take this on if you want)
- Couple questions

.vscode/settings.json
```diff
@@ -1,97 +1,4 @@
...
-        "thread": "cpp",
-        "tuple": "cpp",
-        "typeinfo": "cpp",
-        "variant": "cpp",
-        "__functional_base": "cpp",
-        "functional": "cpp",
-        "pointers": "cpp",
-        "future": "cpp",
-        "multi_span": "cpp"
-    },
```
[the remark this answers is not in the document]
rickwebiii: I have an extension that underlines typos and having that stored here is convenient. If people hate the settings file, I can remove it in another CR. `launch.json` is the important one we do want to share, as it's got all the "how to launch he unit tests so you can debug them" specifications.

sunscreen_math/src/metal_impl/mod.rs
rickwebiii: The real TODO is "is this even sound?" The real issue is that Runtime contains types that have mutable pointers in them, which makes them not automatically `impl Sync+Send`.

I don't know which (if any) Metal types can safely be shared between threads, but I needed to do this to make the POC work. I'll investigate this some more.

Command queues *do* appear to be [thread safe](https://developer.apple.com/library/archive/documentation/Miscellaneous/Conceptual/MetalProgrammingGuide/Cmd-Submiss/Cmd-Submiss.html#//apple_ref/doc/uid/TP40014221-CH3-SW14) so at least that part of the Runtime can be `Sync`.

sunscreen_math/src/metal_impl/mod.rs
rickwebiii: 👍

sunscreen_math/src/metal_impl/ristrettovec.rs
rickwebiii: 10 is the length in `u32`s a Field is in the 32-bit backend. IIRC I made `x` and such's type `Deref` into `[u32; 10]`, so I can call `len()`.

sunscreen_math/src/metal_impl/ristrettovec.rs
rickwebiii: I don't actually know the rules around FFI and uninitialized memory and need to study this a bit more. Accessing uninit data in Rust is UB, but if FFI returned a region of memory it allocated, how would Rust tell the difference? I think I just want to leave the TODO here for now and I'll make an Issue for someone to research and come up with the answer.

sunscreen_math/src/metal_impl/ristrettovec.rs
rickwebiii: UB = undefined behavior. It's a bad scary thing you don't want.

sunscreen_math/src/metal_impl/ristrettovec.rs
rickwebiii: It would be a bit of overhead. You would have to allocate a temporary *vec* of elements to store the negation and run 2 GPU kernels instead of 1.

sunscreen_math/src/metal_impl/ristrettovec.rs
rickwebiii: Yeah, I could make a "vec_binary_op" function that could do this. Not a terrible idea.

sunscreen_math/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+lazy_static = "1.4.0"
+metal = { version = "0.24.0", optional = true }
+rand = "0.8.5"
+curve25519_dalek = { package = "sunscreen_curve25519", path = "../sunscreen_curve25519", features = ["serde", "u64_backend"], default-features = false }
+rayon = "1.6.1"
+
+[features]
+default = []
+nightly-features = []
+metal = ["dep:metal"]
```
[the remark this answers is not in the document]
rickwebiii: Yes, you must explicitly enable the feature and we'll plumb that through to Sunscreen at some point.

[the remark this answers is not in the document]
rickwebiii: I've been thinking about our test matrix a bit and what CI we want to have. A Mac build is likely something we'll want to have, and a WASM build too (it's currently broken). FYI Mac compute costs 10x what Linux does for Github actions, and I don't know if the Docker container has GPU access (needed to run these tests). We don't need a *good* GPU.

sunscreen_math/src/metal_impl/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+const SHADERLIB: &[u8] =
+    include_bytes!(concat!(env!("OUT_DIR"), "/curve25519-dalek.test.metallib"));
+
+pub struct Runtime {
+    device: Device,
+    lib: Library,
+    command_queue: CommandQueue,
+}
+
+/// TODO Remove
```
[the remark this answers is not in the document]
rickwebiii: More like, TODO, figure out if this is even sound...

sunscreen_math/src/metal_impl/ristrettovec.rs
rickwebiii: I even tested this 😄

sunscreen_math/Cargo.toml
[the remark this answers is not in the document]
ryanorendorff: https://blog.pantsbuild.org/m1-support-self-hosted/

If we wanted to set up an ARM Mac runner then this seems like a decent way to do it.

sunscreen_math/src/metal_impl/mod.rs
rickwebiii: I wrapped the non-thread safe things in Mutex and command queue is thread safe. Thus, it is now sound to impl this.

sunscreen_math/Cargo.toml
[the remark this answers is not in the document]
rickwebiii: I'll make an issue for adding mac and WASM CI builds.

sunscreen_math/src/metal_impl/mod.rs
rickwebiii: We have no guidelines around this.

sunscreen_math/src/metal_impl/ristrettovec.rs
rickwebiii: I deleted this remark as it's no longer relevant.

sunscreen_math/src/metal_impl/ristrettovec.rs
rickwebiii: None, it's a simple NewType. That's how you access the one field.

```
struct Foo(Bar)

// my_foo.0 returns the `Bar`
```

sunscreen_math/src/metal_impl/ristrettovec.rs
rickwebiii: It is always true, so the compiler can remove this for us. We run our CI and test in release mode, so we want to be alerted if this is ever not true.

sunscreen_math/src/metal_impl/mod.rs
```diff
@@ -24,13 +26,21 @@ const SHADERLIB: &[u8] = include_bytes!(concat!(
 const SHADERLIB: &[u8] =
     include_bytes!(concat!(env!("OUT_DIR"), "/curve25519-dalek.test.metallib"));
 
+pub struct Grid([(u64, u64); 3]);
```
samtay, line 29: I think this is currently just adding `.0` noise, and probably a type synonym would be fine, but I suppose it doesn't hurt much. Could be useful in the future.

sunscreen_math/src/metal_impl/mod.rs
```diff
@@ -54,18 +64,33 @@ impl Runtime {
...
         self.device
+            .lock()
+            .unwrap()
             .new_buffer(len as u64, MTLResourceOptions::StorageModeShared)
     }
 
-    pub fn run(&self, kernel_name: &'static str, data: &[&Buffer], grid: [(u64, u64); 3]) {
-        println!("{}", self.device.recommended_max_working_set_size());
+    pub fn run(&self, kernel_name: &'static str, data: &[&Buffer], grid: Grid) {
+        println!(
```
samtay, line 73: Do we want to keep this print ?

samtay (approved):
:+1:

sunscreen_math/src/metal_impl/mod.rs
rickwebiii: No, good call

PR #216, opened 2023-02-21
Rweber/zkp

rickwebiii:


sunscreen_math/src/pina.rs
```diff
@@ -0,0 +1,208 @@
...
+    fn mul(self, rhs: PinaScalarVec) -> Self::Output {
+        self * &rhs
+    }
+}
+
+impl Mul<&PinaScalarVec> for &PinaRistrettoPointVec {
+    type Output = PinaRistrettoPointVec;
+
+    fn mul(self, rhs: &PinaScalarVec) -> Self::Output {
+        const GPU_SHARE: f64 = 0.8;
```
rickwebiii, line 99: This constant is a magic number that makes it faster on my machine. The actual PINA algorithm requires you run a microbenchmark for each device type (i.e. CPUs and GPUs), so this is still very much WIP but not the highest priority.

PR #221, opened 2023-03-06
Rweber/zkp

rickwebiii:
Sped up scalar inversion and moved logproof benchmark to criterion and outside lib.

Apparently, u64 multiplication is faster when you don't decompose into lo and hi words yourself.

```
Before:
2082441.6323093667 inversions/s

After:
3158318.637247387 inversions/s
```

Unfortunately, scalar inversion is a small part of the overall runtime, so this gain has negligible impact.


logproof/benches/linear_relation.rs
```diff
@@ -0,0 +1,122 @@
...
+        [zero.clone(), p_1.clone(), zero.clone(), one.clone()],
+        [delta, p_0.clone(), one.clone(), zero.clone()],
+        [zero.clone(), p_1, zero, one],
+    ]);
+
+    // Secret key
+    // a = random in q
+    // e_1 = q / 2p
+    // c_1 = s * a + e_1 + del * m
+    // c_2 = a
```
ryanorendorff, line 65: Do you need the comment twice?

logproof/src/linear_algebra.rs
```diff
@@ -12,13 +12,28 @@ use crate::fields::{FieldFrom, FpRistretto};
...
+/**
+ * An `m x n` matrix of elements.
+ *
+ * # Remarks
+ * Matrix elements can be any type that implements [`Add`], [`Mul`],
+ * [`Clone`], [`Zero`].
+ */
 pub struct Matrix<T>
 where
     T: Zero + Clone,
```
ryanorendorff, line 24: I'm guessing there is no way to add the add and multiply constraint easily?

ryanorendorff (approved):

logproof/src/linear_algebra.rs
rickwebiii: Added clarity. Add and SmartMul aren't strictly required, but they enable more operations.

PR #231, opened 2023-03-23
Parallelize RistrettoVec encoding

rickwebiii:


sunscreen_math/src/metal_impl/ristrettovec.rs
```diff
@@ -45,32 +45,51 @@ impl GpuRistrettoPointVec {
...
-            for (j, w) in x.iter().enumerate() {
-                data_slice[(j + 0 * u29_len) * len + i].write(*w);
-            }
+                for (j, w) in x.iter().enumerate() {
+                    let elem = data_ptr.add((j + 0 * u29_len) * len + i);
 
-            for (j, w) in y.iter().enumerate() {
-                data_slice[(j + 1 * u29_len) * len + i].write(*w);
-            }
+                    (*elem).write(*w);
```
rickwebiii, line 72: Since we're already using pointers, the `MaybeUninit` interface makes less sense here. Writing to a pointer doesn't incur UB because you'll never attempt to drop the object already at that location.

sunscreen_math/Cargo.toml
```diff
@@ -26,7 +26,7 @@ ocl = { version = "0.19.4", optional = true }
 criterion = "0.4.0"
 
 [features]
-default = []
+default = ["metal"]
```
rickwebiii, line 29: Remove

sunscreen_math/src/metal_impl/ristrettovec.rs
```diff
@@ -45,32 +45,51 @@ impl GpuRistrettoPointVec {
...
+                let data_ptr = data_ptr as *mut MaybeUninit<u32>;
 
-            let u29_len = x.len();
+                let u29_len = x.len();
 
-            for (j, w) in x.iter().enumerate() {
-                data_slice[(j + 0 * u29_len) * len + i].write(*w);
-            }
+                for (j, w) in x.iter().enumerate() {
+                    let elem = data_ptr.add((j + 0 * u29_len) * len + i);
```
samtay, line 70: Huh, what are these `0 * u29_len` values? Looking at the types, it looks like `u29_len` is always `10: usize`, right? If so, why multiply it by zero?

samtay: Oh, perhaps the next ones should still increment the offset like they used to? I.e. multiply by 0, 1, 2, 3 for x, y, z, t respectively?

sunscreen_math/src/metal_impl/ristrettovec.rs
rickwebiii: Let's not mess with that.

sunscreen_math/src/metal_impl/ristrettovec.rs
rickwebiii: Exactly. The offsets aren't zero for y, z, t and this just makes the alignment be pretty.

samtay: Gotcha. Just to be clear though, in the current diff, the offsets are currently `0` for all of `x, y, z, t`.

rickwebiii: Doh

PR #232, opened 2023-03-29
Allow library consumers to build gadgets

samtay:
- Expose `ZkpContextOps` . This is necessary for people to be able to build their own gadgets.
- Glob `examples/*` as workspace members, why not


rickwebiii (approved):
🚢

PR #234, opened 2023-04-01
Rweber/perf

rickwebiii:
Instead of recursively compacting the generator vectors using scalar multiplications, we precompute what the factors should be and perform a single MSM. See Bulletproofs section 3.1 for a more formal writeup of this optimization.

### DIFF
--- a/logproof/src/inner_product.rs
+++ b/logproof/src/inner_product.rs
@@ -1,3 +1,5 @@
+use std::time::Instant;
+
 use curve25519_dalek::{
     ristretto::{CompressedRistretto, RistrettoPoint},
     scalar::Scalar,
@@ -5,10 +7,11 @@ use curve25519_dalek::{
 use digest::ExtendableOutput;
 use digest::XofReader;
 use merlin::Transcript;
+use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use sha3::{self, digest::Update, Shake256};
 
-use sunscreen_math::RistrettoPointVec;
+use sunscreen_math::{RistrettoPointVec, ScalarVec};
 
 use crate::error::ProofError;
 use crate::{linear_algebra::InnerProduct, math::rand256};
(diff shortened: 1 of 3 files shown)


logproof/src/inner_product.rs
```diff
@@ -377,25 +392,53 @@ impl InnerProductProof {
...
-            let (h_t, h_b) = h.split_at(n_2);
+        let mut c = vec![];
 
-            let t_1 = t_1.decompress().ok_or(ProofError::MalformedProof)?;
-            let t_minus1 = t_minus1.decompress().ok_or(ProofError::MalformedProof)?;
+        // See Bulletproofs paper section 3.1 for what this optimization is. We're deferring
+        // folding our generators g and h and instead computing factors s from each of the
+        // challenge scalars.
+        //
+        // This allows us to compute a single MSM at the end to compute g and h rather that
```
rickwebiii, line 403: *than

logproof/src/inner_product.rs
```diff
@@ -291,6 +305,9 @@ impl InnerProductProof {
         let lhs = t_pprime * c + w + w_prime * c_inv;
         let rhs = g * self.z_1 + h * self.z_2 + a * (c_inv * self.z_1 * self.z_2) + u * self.tau;
 
+        println!("Post fold {}s", now.elapsed().as_secs_f64());
```
rickwebiii, line 308: Imma keep this logging for now...

logproof/src/inner_product.rs
```diff
@@ -655,5 +701,76 @@ mod tests {
...
+
+            let s_i = |i| {
+                c.iter().rev().enumerate().fold(Scalar::one(), |p, (j, x)| {
+                    if i & (0x1 << j) != 0 {
+                        p * x
+                    } else {
+                        p
+                    }
+                })
+            };
```
samtay, line 751: We should probably factor out the common parts that are duplicated here, so that the test is always testing the actual implementation.

rickwebiii: This test can actually be removed. It was a TDD thing I made to test that the algorithm for computing the s values worked. We now know the algorithm works because verification tests still pass.

PR #236, opened 2023-04-22
Add a `bfv::Unsigned` type

samtay:
This should be cleaned up and have some tests in `./sunscreen/tests`. Opening for visibility on initial approach / representation.

Some notable decisions:

1. We just wrap the crypto big ints. If desirable we could add macros to define synonyms and to/from impls for all the `UInt`s that that crate defines. (Also we should probably add some macros for all the boilerplate arithmetic impls before merging).
2. We use wrapping arithmetic, as the current `bfv::Signed` type does (in release mode anyway).

### DIFF
--- a/sunscreen/src/types/bfv/unsigned.rs
+++ b/sunscreen/src/types/bfv/unsigned.rs
@@ -1,0 +1,541 @@
+use crypto_bigint::{UInt, Wrapping};
+use seal_fhe::Plaintext as SealPlaintext;
+
+use crate as sunscreen;
+use crate::{
+    fhe::{with_fhe_ctx, FheContextOps},
+    types::{
+        ops::{
+            GraphCipherAdd, GraphCipherConstAdd, GraphCipherConstMul, GraphCipherConstSub,
+            GraphCipherMul, GraphCipherPlainAdd, GraphCipherPlainMul, GraphCipherPlainSub,
+            GraphCipherSub, GraphConstCipherSub, GraphPlainCipherSub,
+        },
+        Cipher,
+    },
+};
+use crate::{
+    types::{intern::FheProgramNode, BfvType, FheType, TypeNameInstance},
+    FheProgramInputTrait, Params, TypeName as DeriveTypeName, WithContext,
+};
+
+use sunscreen_runtime::{
+    InnerPlaintext, NumCiphertexts, Plaintext, TryFromPlaintext, TryIntoPlaintext,
+};
+
+use std::ops::*;
+
+// TODO lil macro to gen all the ones from crypto_bigint
+// TODO tests in sunscreen/tests dir
+
+/// Unsigned 64 bit integer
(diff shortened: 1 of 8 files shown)


sunscreen/src/types/bfv/unsigned.rs
```diff
@@ -0,0 +1,358 @@
...
+}
+fn wrapping_sub<const LIMBS: usize>(lhs: UInt<LIMBS>, rhs: UInt<LIMBS>) -> UInt<LIMBS> {
+    (Wrapping(lhs) - Wrapping(rhs)).0
+}
+
+macro_rules! impl_std_op {
+    ($($op:ident),+) => {
+        $(
+            paste! {
+                impl<const LIMBS: usize> $op for Unsigned<LIMBS> {
```
rickwebiii, line 166: I never did figure out how to wrote macro_rules! and had to use proc_macro, which entails a ton of shittiness.

A thing I've always wanted was a macro that would imlement each combination of borrow/owned for lhs and rhs in std ops. I think this would even compose well with such a macro!

Nothing to do here.

rickwebiii (approved):

sunscreen/src/types/bfv/unsigned.rs
```diff
@@ -0,0 +1,358 @@
...
+        plaintext: &Plaintext,
+        params: &Params,
+    ) -> std::result::Result<Self, sunscreen_runtime::Error> {
+        let val = match &plaintext.inner {
+            InnerPlaintext::Seal(p) => {
+                if p.len() != 1 {
+                    return Err(sunscreen_runtime::Error::IncorrectCiphertextCount);
+                }
+
+                let bits = usize::min(std::mem::size_of::<UInt<LIMBS>>() * 8, p[0].len());
```
rickwebiii, line 101: plaintext values have 1024 to 32768 bits of precision depending on the scheme parameters, and taking the min here means you're effectively decoding values modulo 2^(64*LIMBS). I'm wondering if cases where the plaintext value is too large to fit in the specified limb it might be better to outright fail? We'll want to document this behavior in the user guide and eventually™️ make a cohesive and similar set of behaviors for the signed type (for which we'll need to make a signed `Int<LIMBS>` native type). Maybe that type will just wrap `UInt<Limbs>` and have a sign flag?

samtay: I might be misunderstanding you, but I don't think we have the problem of plaintext values fitting into a particular limb. The output is a sum, coefficient by coefficient, of values like:
```
(UInt::from_u8(0x1) << i) * (UInt::from_u64(coeff))
```
so we don't build this up limb by limb (like the original version did when I initially opened the PR).

As I reread your comment, I'm thinking you mean "too large to fit into the specified _number of_ limbs". I.e. throw an error when `std::mem::size_of::<UInt<LIMBS>>() * 8 < p[0].len()`? If so, that does seem reasonable to me.

PR #237, opened 2023-05-01
Make gadgets sync

samtay:
This lets Application types be send/sync, which is useful for lazy/static definitions.


rickwebiii (commented):
👍

PR #239, opened 2023-05-12
Separate OpenCL shaders into multiple files

rickwebiii:
Minor change: move serialization tests to sunscreen to break a dependency cycle.
Main change:
Allow OpenCL shaders to be split across multiple files. We previously had this restriction for 2 reasons:
* Apple's OpenCL compiler is kinda crappy. For example, if you passed 2 source files that each declared
```c
typedef struct { } Foo;
```

it would complain about redeclaration, even though they should be in different compilation units.

* How do we distribute header files to end users and manifest them onto disk so we can pass some "-I include_dir" to the OpenCL compiler? We compile shaders at runtime because we don't know in advance what platform and hardware the end user has.

To work around this, we have a `main.cl` that `#include`s all the other `.cl` files and invoke Clang to preprocess `main.cl` to emit a single giant source file. Thus, developers get the benefit of code separation, but OpenCL sees a single giant file as before. We bake this giant file into the binary with `include_str`.


sunscreen_math/build.rs
```diff
@@ -306,6 +357,18 @@ fn compile_metal_shaders() {
     }
 }
 
+fn validate_command_output(output: Output, panic_msg: &str) {
```
rickwebiii, line 360: I should probably update all the other users...

PR #243, opened 2023-05-25
Fix 64-bit pointer assumptions

samtay:
We'll still need the `syn`/`proc-macro2` fixes to get `build --target wasm32-unknown-emscripten` working. This fixes all the other issues.

We still have some issues for wasm when gated behind e.g. `test`, but I'm punting on that, just wanted to get these in for now.

### DIFF
--- a/benchmarks/bfv_zkp/src/bfv.rs
+++ b/benchmarks/bfv_zkp/src/bfv.rs
@@ -263,13 +263,13 @@ fn ark_bigint_to_native_field<B: ZkpBackend, F: MontConfig<N>, const N: usize>(
     assert!(N <= 8);
 
     let x = x.into_bigint();
-    let mut out = ZkpBigInt::ZERO;
+    let mut words = [0; 8];
 
-    for (i, l) in x.0.iter().enumerate() {
-        out.0.limbs_mut()[i] = crypto_bigint::Limb(*l);
+    for i in 0..N {
+        words[i] = x.0[i];
     }
 
-    NativeField::from(out)
+    NativeField::from(ZkpBigInt::from_words(words))
 }
 
 type BpBackendField = <BulletproofsBackend as ZkpBackend>::Field;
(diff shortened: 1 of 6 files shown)


benchmarks/bfv_zkp/src/bfv.rs
```diff
@@ -260,16 +260,9 @@ pub fn compile_proof() -> ZkpApplication {
 fn ark_bigint_to_native_field<B: ZkpBackend, F: MontConfig<N>, const N: usize>(
     x: Fp<MontBackend<F, N>, N>,
 ) -> NativeField<B::Field> {
-    assert!(N <= 8);
```
rickwebiii, line 263: We should keep some assertion here, as converting a bigint is only correct if the Arkworks int <= 512 bit, regardless of the limb size.

Maybe:
```
let max_limbs = if cfg!("32_bit") {
  16
} else if cfg!("64_bit) {
  8
} else {
  unreachable!();
}

assert!(N <= max_limbs);
```

and then somewhere else doing
```
#[cfg(not(any("32_bit", "64_bit"))))]
compile_error!("Unsupported limb width");
```

samtay, line 263: @rickwebiii idk where your comment went, but I see it in my email notifications :shrug: will quote the relevant parts here:

> We should keep some assertion here, as converting a bigint is only correct if the Arkworks int <= 512 bit, regardless of the limb size.

This assert `N <= 8` still happens, it just happens in `ark_bigint_to_zkp_bigint`. (This code was doing the same thing so I just called the existing function.)

> Maybe:
> ```
> let max_limbs = if cfg!("32_bit") {
>   16
> } else if cfg!("64_bit) {
>   8
> } else {
>   unreachable!();
> }
> ```

We actually don't need to worry about that here, because these `N`s are used in `BigInt<N>`s which are always `[u64; N]` regardless of `target_pointer_width`. So, `N <= 8` iff `bitsize <= 512`

PR #244, opened 2023-05-25
Use https in submodule remotes

samtay:
This makes specifying our git repo as a cargo dependency much less painful.

We could further optimize this by specifying `update = none` (see https://github.com/rust-lang/cargo/pull/10717) on the projects unrelated to the sunscreen library (i.e. mdBook, rust-playground). Don't know if that is worth it or not.


rickwebiii (approved):
🍖

PR #252, opened 2023-06-07
Fix broken doc links

samtay:
1. Fix broken links in the mdbook documentation
2. Fix broken api docs + amend CI to catch these issues.


sunscreen_docs/src/fhe_programs/types/cipher.md
samtay, line 1: This snippet already lives in the `types.md`, and this file is not specified in the `summary.md`, so it doesn't exist as a page in the site output. It was probably just leftover when reorganizing stuff.

PR #253, opened 2023-06-08
Added Sudoku ZKP Example

bryanyli:
Added a example ZKP program that allows a prover to prove that they have a valid solution to a given sudoku puzzle.

### DIFF
--- a/examples/sudoku_zkp/src/main.rs
+++ b/examples/sudoku_zkp/src/main.rs
@@ -152,7 +152,7 @@ mod tests {
 
         let proof = runtime.prove(prog, cons.clone(), vec![], board).unwrap();
 
-        let verify = runtime.verify(prog, &proof, cons.clone(), vec![]);
+        let verify = runtime.verify(prog, &proof, cons, vec![]);
 
         assert!(verify.is_ok());
     }
@@ -197,7 +197,7 @@ mod tests {
 
         let cons: Vec<ZkpProgramInput> = vec![ex_puzzle.map(|a| a.map(BPField::from)).into()];
 
-        let proof = runtime.prove(prog, cons.clone(), vec![], board);
+        let proof = runtime.prove(prog, cons, vec![], board);
 
         assert!(proof.is_err());
     }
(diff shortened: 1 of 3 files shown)


examples/sudoku_zkp/src/main.rs
```diff
@@ -0,0 +1,254 @@
...
+    let one = NativeField::<F>::from(1).into_program_node();
+    let nine = NativeField::<F>::from(9).into_program_node();
+    let zero = NativeField::<F>::from(0).into_program_node();
+
+    for i in 0..9 {
+        for j in 0..9 {
+            let square = board[i][j].into_program_node();
+            let constraint = constraints[i][j].into_program_node();
+            square.constrain_ge_bounded(one, 8);
+            square.constrain_le_bounded(nine, 8);
```
samtay, line 80: It occurs to me that since we are ensuring the sets are `{1, ... , 9}` via that polynomial from `contains_circuit`, I don't think we need these two bounds checks ensuring `1 <= square <= 9`.

ryanorendorff:
It may be helpful for an example to include a few more type annotations; just a suggestion though.

examples/sudoku_zkp/src/main.rs
rickwebiii: You don't.

examples/sudoku_zkp/src/main.rs
```diff
@@ -0,0 +1,254 @@
...
+    #[constant] constraints: [[NativeField<F>; 9]; 9],
+    board: [[NativeField<F>; 9]; 9],
+) {
+    fn contains_circuit<F: BackendField>(
+        arr: [ProgramNode<NativeField<F>>; 9],
+        i: NativeField<F>,
+    ) -> ProgramNode<NativeField<F>> {
+        let mut circuit = NativeField::<F>::from(1).into_program_node();
+        for a in arr {
+            circuit = circuit * (i.into_program_node() - a);
```
rickwebiii, line 66: I would probably do the `for i in 1..=9` iteration in this function. Rename this helper `assert_unique_numbers` and add a comment that this function checks that the values 1 through 9 each appear once in `arr`.

rickwebiii: And add the `.constrain_eq(0)` in here.

rickwebiii (approved):
Fix the 2 comments, then :shipit:

samtay (approved):
looks great!

PR #263, opened 2023-06-26
commented out compile errors, added group_id to context

matthew-liu801:
commented out some code from main, bulid in debugger file to fix compile errors. also added optional field group_id for context


sunscreen_compiler_common/src/context.rs
```diff
@@ -229,9 +236,24 @@ where
...
+        {
+            Self {
+                graph: CompilationResult::<O>::new(),
+                data,
+            }
+        }
+        
+        #[cfg(feature = "debugger")]
+        {
+            static GROUP_ID: AtomicU64 = AtomicU64::new(0);
```
rickwebiii, line 249: You shouldn't need atomics; a simple counter should suffice. `FheProgramContext` isn't thread-safe anyways.

rickwebiii (approved):

PR #264, opened 2023-06-26
Build to_2s_complement by individual elements.

ryanorendorff:
This change is a no-op on the result; it is a precursor to having bounds that can vary on each element in the input array.

### DIFF
--- a/logproof/src/linear_relation.rs
+++ b/logproof/src/linear_relation.rs
@@ -834,12 +834,64 @@ impl LogProof {
     }
 
     /**
+     * Takes an individual values in a field `Zq`, treats the value as signed [q's
+     * complement](https://en.wikipedia.org/wiki/Method_of_complements) and
+     * converts the value to binary 2's complement. This is then appended to an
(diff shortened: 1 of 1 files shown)


logproof/src/linear_relation.rs
```diff
@@ -833,13 +833,65 @@ impl LogProof {
...
+     * converts the value to binary 2's complement. This is then appended to an
+     * provided BitVec.
+     *
+     * `value` is the element in Zq and `b` is the number of bits needed
+     * to represent the signed value.
+     *
+     * This modifies bitvec in place.
+     *
+     */
+    #[inline]
```
samtay, line 848: Don't know that this needs to be inline but :shrug:

samtay (approved):

logproof/src/linear_relation.rs
ryanorendorff: Ya I was trying to keep the exact same setup as before but it doesn't need to be inlined. Removed.

PR #265, opened 2023-06-26
Misc typos / Fix lib exports

samtay:
- Drive-by fixes to rust docs
- I missed this in Bryan's initial sudoku PR, but he had to specify a dependency on the local `sunscreen_zkp_backend` crate; changes here allow users to import those types from the main `sunscreen` lib.


samtay:
Going to go ahead and merge to unblock zkp docs progress. Can address any comments in a follow up.

PR #267, opened 2023-06-27
Mliu backend

matthew-liu801:
many changes for adding group id's to fheprogramnodes if "debugger" feature is enabled


sunscreen_compiler_common/src/context.rs
```diff
@@ -31,7 +37,11 @@ where
      * Creates a new [`NodeInfo`].
      */
     pub fn new(operation: O) -> Self {
```
rickwebiii, line 39: Oh, this will be a problem. You'll need to change this signature to `operation: O, group_id: usize` when the debugger feature is enabled. That will give you a bunch more errors to go fix.

sunscreen_compiler_common/src/context.rs
```diff
@@ -258,6 +268,17 @@ where
      * Add a node to the parse graph.
      */
     pub fn add_node(&mut self, operation: O) -> NodeIndex {
+        #[cfg(feature = "debugger")]
+        {
+            let group_id = self.group_counter;
+            self.group_counter += 1;
```
rickwebiii, line 274: Probably don't increase the group counter each time you make a node, but rather when you push a group onto the stack that doesn't already exist in your group data structure.

rickwebiii (approved):

PR #274, opened 2023-07-05
Add CSV stdout for logproof bench

ryanorendorff:
Makes a few small improvements to logproof benchmarks to make them easier to run:

1. Outputs the results at the end as a csv formatted table for easier plotting in other programs.
2. Makes the optional thermal throttle a boolean instead of commenting out code. This change was because commenting out the code also required adding an import (`Duration`), so this is slightly easier to use.


samtay (approved):
We should really just use criterion :laughing: but in the meantime this looks good.

PR #276, opened 2023-07-06
Mixed bounds in logproof

ryanorendorff:
Adds individual bounds to each coefficient in the short discrete log proof (SDLP) implementation.

This enables some small performance increase for smaller proofs (up to approximately 75%). Additionally this enables the prover and verifier to agree on which components may or may not be interesting by specifying a bound of zero on a coefficient.

This change somewhat abuses the notation for `Matrix`, as it enables essentially `Matrix<Vec<u64>>` to be written in order to keep track of the bounds. We can choose another way to store the bounds; the advantage of the current method is that it matches the shape and serialization method as the `s` matrix as required to implement the bounds.

Some pieces are put under the `LogProof` module as they are only notations relevant to the SDLP notation, but they could be moved elsewhere. Examples include `serialize_bounds` and `serialize_bounds_twos_complement_coefficients`.

Negative coefficients are also now used in the transcripts match tests to ensure that the mixed bounds work with negative values.

Some pieces that can be added, either in this PR or in another PR:

1. A (set of) helper function(s) to enable creating bounds more easily. For example, constant bounds, constant bounds per term, tight bounds based on a matrix of the maximum coefficient values in `s`, etc.

The tests for `k > 1` only pass when #275 is merged, hence this PR is stacked on top of #275.

Changes in the algorithm are shown in red in the following protocol diagram:
sdlp_changes


logproof/src/linear_algebra.rs
```diff
@@ -350,6 +350,45 @@ where
...
+        }
+
+        Self {
+            rows,
+            cols,
+            data: data.to_owned(),
+        }
+    }
+}
+
```
ryanorendorff, line 391: There is no change here, I simply moved all the `impl From... for Matrix<F>` to the same block of code.

logproof/src/linear_relation.rs
```diff
@@ -24,6 +24,22 @@ use crate::{
...
+type Bounds = Vec<u64>;
+
+impl Zero for Bounds {
+    // The empty vector could be seen as no bounds. Also follows the field
+    // properties.  Although realistically this would be indexed by the
+    // dimension d.
+    fn zero() -> Self {
+        Vec::new()
+    }
+}
```
ryanorendorff, line 36: I don't love this implementation. I mostly wanted to use a matrix like container type so I can check that all the dimensions of `s` matched the dimensions of the bounds. We could add a new type for it, but I was attempting to reduce the number of new largely equivalent container types we had.

logproof/src/linear_relation.rs
```diff
@@ -1000,18 +1157,56 @@ mod test {
...
+                                    } else {
+                                        next_higher_power_of_two(x)
+                                    }
+                                })
+                                .collect::<Vec<u64>>()
+                        })
+                        .collect::<Vec<Vec<u64>>>()
+                })
+                .collect::<Vec<Vec<Vec<u64>>>>(),
+        );
```
ryanorendorff, line 1188: This seems bananas in terms of layers. If desired I could look into more comprehensible alternatives. Although 3 layer vecs may always look a bit complex.

logproof/src/linear_relation.rs
```diff
@@ -950,20 +1103,24 @@ mod test {
...
+        Q: Field + Zero + Clone + FftField,
+    {
+        a: MatrixPoly<Q>,
+        s: MatrixPoly<Q>,
+        t: MatrixPoly<Q>,
+        f: DensePolynomial<Q>,
+        b: Matrix<Bounds>,
+    }
+
+    fn test_lattice<Q>(k: usize) -> LatticeProblem<Q>
```
ryanorendorff, line 1123: Added this to make clippy happy with the complexity of the return type.

logproof/src/linear_relation.rs
```diff
@@ -641,6 +700,30 @@ impl LogProof {
...
+     * and flattens the result. If the each vector happens to be the same
+     * length, then this is the same as the following operation:
+     * `vec(diag(v) * M)` where `vec` is row major vectorization.
+     *
+     * - v: Vector to diagonalize
+     * - m: Matrix as an array of vectors
+     *
+     * Note: the elements in `m` do not need to be the same size.
+     */
+    pub(crate) fn scale_rows_and_flatten<T>(v: &[T], m: &[Vec<T>]) -> Vec<T>
```
ryanorendorff, line 714: It's not the most amazing name since each row can be a different length but it was the best I could come up with. 🤷

logproof/benches/linear_relation.rs
```diff
@@ -216,7 +233,7 @@ criterion_group!(
     params_4096_1ct,
     params_4096_2ct,
     params_4096_3ct,
-    print_results
+    printout
```
ryanorendorff, line 236: Change back to print_results

logproof/src/linear_relation.rs
```diff
@@ -227,14 +284,14 @@ where
         a: &MatrixPoly<Q>,
         s: &MatrixPoly<Q>,
         t: &MatrixPoly<Q>,
-        bound: u64,
+        bounds: Matrix<Bounds>,
```
ryanorendorff, line 287: Maybe change this also to a reference

logproof/src/math.rs
```diff
@@ -264,10 +279,20 @@ pub fn print_polynomial<F: Field>(p: &DensePolynomial<F>) {
...
-    let coeffs = coeffs.iter().map(|x| F::from(*x)).collect();
+pub fn make_poly<F: Field + From<u64>>(coeffs: &[i64]) -> DensePolynomial<F> {
+    let zero = F::zero();
+    let coeffs = coeffs
+        .iter()
+        .map(|x| {
+            if *x >= 0 {
+                F::from(*x as u64)
+            } else {
+                zero - F::from(-*x as u64)
```
ryanorendorff, line 292: Change to abs

logproof/src/linear_relation.rs
ryanorendorff: We have determined this is ok for now :)

logproof/src/linear_relation.rs
ryanorendorff: Also determined this was ok since it is test code.

PR #278, opened 2023-07-10
Properly factor inserting literal plaintexts

samtay:
This came up while doing unrelated ZKP stuff. Not sure why I didn't do this in the initial PR.


ryanorendorff (approved):
LGTM!

PR #279, opened 2023-07-11
ZKP tweaks

samtay:
Some changes made while writing zkp user docs:

- sudoku cleanup
- new zkp serialization example
- exports necessary to use zkp compiler
- convenience exports (e.g. `BulletproofsField`)
- couple of doc fixes
- rename `Gadget::compute_inputs` to `Gadget::compute_hidden_inputs`


.github/workflows/rust.yml
```diff
@@ -42,26 +42,10 @@ jobs:
       run: cargo fmt --check
     # Build and run the tests
     - name: Build and run tests
-      run: cargo test --verbose --release
-    # Run our non-interactive examples and assert the complete without error
-    - name: Verify examples (amm)
```
rickwebiii, line 47: We're no longer running our examples?

rickwebiii (approved):

.github/workflows/rust.yml
samtay: The `test` step runs all tests in the workspace, including examples. I just made sure each example had a `#[test]` testing `main`.

PR #285, opened 2023-07-24
Export components used in encryption, prove encryption is valid as ZKP

ryanorendorff:
Progress towards exporting the components from an encryption. Putting this up now in case anyone has comments on the implementation.


logproof/src/fields.rs
```diff
@@ -48,10 +48,15 @@ where
...
+ *
+ * This can be derived by running
+ * `CoefficientModulus::bfv_default(8192, SecurityLevel::TC128)`
+ * or by running the underlying SEAL function
+ * `CoeffModulus::BFVDefault(8192, sec_level_type::tc128)`
  */
 #[derive(MontConfig)]
-#[modulus = "421249101157150430150591791601812858371395928330411389778873040897"]
 #[generator = "3"]
+#[modulus = "421249101157150430150591791601812858371395928330411389778873040897"]
```
rickwebiii, line 59: Hmm... do we use the last modulus or not without the special modulus? You can take a ciphertext and count that you should have 2 polynomials, each with 8192*{4 | 5} RNS lanes. That will tell you whether this is the right modulus or not.

logproof/src/fields.rs
```diff
@@ -62,6 +67,11 @@ pub struct SealQ128_8192 {}
  *
  * SEAL uses Q =
  * 0xffffee001, 0xffffc4001, 0x1ffffe0001
+ *
+ * This can be derived by running
+ * `CoefficientModulus::bfv_default(4096, SecurityLevel::TC128)`
+ * or by running the underlying SEAL function
+ * `CoeffModulus::BFVDefault(4096, sec_level_type::tc128)`
```
rickwebiii, line 74: Ditto do we use 2 or 3 modulii?

logproof/src/fields.rs
```diff
@@ -70,16 +80,20 @@ pub struct SealQ128_4096 {}
...
+ *
+ * This can be derived by running
+ * `CoefficientModulus::bfv_default(2048, SecurityLevel::TC128)`
+ * or by running the underlying SEAL function
+ * `CoeffModulus::BFVDefault(2048, sec_level_type::tc128)`
  */
 #[derive(MontConfig)]
 #[generator = "3"]
-#[modulus = "4611686014132420865"]
+#[modulus = "18014398492704769"]
```
rickwebiii, line 96: Hmm... 2048 and 1024 shouldn't be affected by the special modulus, so not sure why this changed.

logproof/src/linear_relation.rs
```diff
@@ -1404,4 +1409,256 @@ mod test {
     fn transcripts_match_k_4() {
         transcripts_match(4);
     }
+
+    fn convert_to_smallint(
```
rickwebiii, line 1413: I don't quite understand what this function is doing. Subtracting modulus values is an odd thing to do. Maybe a code walkthrough tomorrow?

logproof/src/linear_relation.rs
```diff
@@ -1404,4 +1409,256 @@ mod test {
...
+        let coeffs2 = coeffs2.into_iter().map(|x| Q::zero() - x).collect();
+
+        DensePolynomial { coeffs: coeffs1 } + DensePolynomial { coeffs: coeffs2 }
+    }
+
+    #[test]
+    fn full_knowledge_bfv_proof() {
+        type Fq = FqSeal128_1024;
+
+        let degree = 1024;
```
rickwebiii, line 1523: 1024 is a good litmus test for you get the right answer because SEAL doesn't use the special modulus for this lattice dimension (or 2048) anyways, so your changes have no effect.

logproof/src/fields.rs
ryanorendorff: I think we do not, since there are 4 RNS lanes in the data. The last modulus is not used in the data itself. So I think we will need to change this number, but I have been focusing on the case where the RNS lanes is 1 so that I can ensure that the full knowledge proof is correct.

logproof/src/fields.rs
ryanorendorff: For 4096 the number specified here is from the key modulus.

logproof/src/fields.rs
ryanorendorff: The numbers just didn't match what was in the file; I am not sure why. The encrypted data also does not exceed the new modulus specified here, so I assume it was just a typo somewhere.

logproof/src/linear_relation.rs
ryanorendorff: Precisely!

logproof/src/linear_relation.rs
ryanorendorff: Sure. Essentially for small negative numbers, the SEAL code that generates the RNS values essentially subtracts that small negative number from each moduli. So the idea here is just to convert back to the original small integers that make up `u` and `e` in order to  be able to encode them in any field (including a higher field) more easily.

logproof/src/fields.rs
ryanorendorff: It seems that both values had an extra set of `0x01` appended:

```
degree 1024        2048
old    0x7e0000101 0x3fffffff00000101
new    0x7e00001   0x3fffffff000001
```

logproof/src/linear_relation.rs
ryanorendorff: I've also added a temporary check that the small int conversion does not change the calculated answer. That test passes.

The main reason for it was to see if I was exporting values that made sense (ie were either ternary or were from the CBD with a std of 3.2)

logproof/src/fields.rs
```diff
@@ -48,10 +48,15 @@ where
...
+ *
+ * This can be derived by running
+ * `CoefficientModulus::bfv_default(8192, SecurityLevel::TC128)`
+ * or by running the underlying SEAL function
+ * `CoeffModulus::BFVDefault(8192, sec_level_type::tc128)`
  */
 #[derive(MontConfig)]
-#[modulus = "421249101157150430150591791601812858371395928330411389778873040897"]
 #[generator = "3"]
+#[modulus = "23945240908173643396739775218143152511335532357255169"]
```
ryanorendorff, line 59: The higher two moduli (4096 and 8192) had multiple components (say N), of which the encryption is done over the product of only the first N - 1 moduli. The final special modulus is only used for modulus switching, which reduces initial noise on encryption in some schemes. 

See https://crypto.stackexchange.com/a/87971

logproof/src/fields.rs
ryanorendorff: Confirmed by saving the modulus value and performing the encryption manually that these new moduli are the correct ones (for 1024 and 2048).

logproof/src/fields.rs
```diff
@@ -105,7 +125,7 @@ pub struct SealQ128_1024 {}
  * Fp expects the modulus to be prime, but ours isn't. We need to be good
  * children and not use anything that relies on field primality.
  */
-pub type FqSeal128_8192 = Fp256<MontBackend<SealQ128_8192, 4>>;
+pub type FqSeal128_8192 = Fp192<MontBackend<SealQ128_8192, 3>>;
```
ryanorendorff, line 128: SEAL's multiprecision uses the same number of limbs as the number of components in the coefficient modulus, even if that number of limbs higher than required to represent the coefficients. In this case, SEAL will use 4 limbs while the coefficient modulus can fit inside 3 limbs.

logproof/src/linear_relation.rs
```diff
@@ -1404,4 +1409,329 @@ mod test {
...
+    }
+
+    /**
+     * Converts a polynomial known to have coefficients less than all of the
+     * moduli in its associated modulus set into regular integers. The main
+     * advantage here over using a polynomial in its normal field is that the
+     * polynomial can be moved to a new field without modulus switching.
+     */
+    #[allow(unused)]
+    fn convert_to_smallint(
```
ryanorendorff, line 1436: These small int functions are not used anymore but I have left them in case we need them later. The original idea was that I wanted to test that the small integer representation, which can be easier to work with in some cases, was equivalent to the full RNS representation. These may be useful in later computations where we are performing calculations in different fields from the field SEAL is using.

logproof/src/math.rs
```diff
@@ -653,6 +653,49 @@ where
     }
 }
 
+/**
+ * Computes the quotient and remainder of a / b.
+ */
+pub fn div_rem_bigint<const N: usize>(a: BigInt<N>, b: BigInt<N>) -> (BigInt<N>, BigInt<N>) {
```
ryanorendorff, line 659: We have multiple math libraries. Currently this is defined in the zkp crate; @rickwebiii what do you think about moving this to `sunscreen_math`? That dependency is already imported into this crate.

sunscreen_runtime/src/runtime.rs
```diff
@@ -392,20 +428,96 @@ where
...
+                            };
+
+                            let r = Plaintext {
+                                data_type: P::type_name(),
+                                inner: InnerPlaintext::Seal(vec![r_context]),
+                            };
+
+                            us.push(u);
+                            es.push(e);
+                            rs.push(r);
```
ryanorendorff, line 514: This could be converted from `Vec<Plaintext>` to just `Plaintext` to match how the Ciphertext is bundled up. I'm not sure that is better: consistency with how the other components are defined or consistency with the ciphertext structure.

sunscreen_runtime/src/runtime.rs
```diff
@@ -392,20 +428,96 @@ where
...
+     * Encrypts the given [`FheType`](crate::FheType) using the given public
+     * key, returning the encrypted value along with the components added to the
+     * message. See `BFVEncryptionComponents` for the pieces returned.
+     *
+     * Note that this will disable the special modulus!
+     *
+     * Returns [`Error::ParameterMismatch`] if the plaintext is incompatible with this runtime's
+     * scheme.
+     */
+    fn encrypt_return_components_switched<P>(
```
ryanorendorff, line 466: The result of this is BFV specific. I believe that is ok for now, but we may want to have a more generic output if we support other backends.

seal_fhe/src/encryptor_decryptor.rs
```diff
@@ -115,6 +119,46 @@ impl Encryptor {
...
+     * The encryption parameters for the resulting ciphertext correspond to:
+     * 1) in BFV, the highest (data) level in the modulus switching chain,
+     * 2) in CKKS, the encryption parameters of the plaintext.
+     * Dynamic memory allocations in the process are allocated from the memory
+     * pool pointed to by the given MemoryPoolHandle.
+     *
+     * * `plainext` - The plaintext to encrypt.
+     * * `disable_special_modulus` - Whether to disable the special modulus switching
+     */
+    pub fn encrypt_return_components(
```
rickwebiii, line 138: `encrypt_return_noise` is probably a better name

logproof/src/math.rs
rickwebiii: I think we should mark this deprecated and use `https://docs.rs/crypto-bigint/latest/crypto_bigint/struct.Uint.html#method.div_rem`

logproof/src/linear_relation.rs
```diff
@@ -1404,4 +1409,402 @@ mod test {
...
+
+    fn full_knowledge_proof<F, const N: usize>(
+        degree: u64,
+    ) -> LatticeProblem<Fp<MontBackend<F, N>, N>>
+    where
+        F: MontConfig<N>,
+    {
+        let degree = degree;
+
+        let plain_modulus = PlainModulus::raw(1032193).unwrap();
```
rickwebiii, line 1546: I think this is too large to N=1024. Maybe we should drop this to like 4096 or some plain modulus that works for N=1024.

Also, might be work moving this to `tests/seal_logproof.rs` and Cargo should pick it up and use it as an E2E test. This file is already enormous.

seal_fhe/src/encryptor_decryptor.rs
ryanorendorff: Well r is not noise; it is the component that handles that seal rounds the message instead of just flooring it.

logproof/src/linear_relation.rs
ryanorendorff: I'll move it.

Turns out `12289` as a plaintext modulus for polynomial modulus degree 1024.

logproof/src/math.rs
ryanorendorff: Done.

logproof/src/linear_relation.rs
ryanorendorff: Marking as resolved. While batching is still used, a smaller modulus was found that worked for each specific encryption dimension and all tests pass.

logproof/tests/seal.rs
```diff
@@ -0,0 +1,293 @@
...
+    linear_algebra::Matrix,
+    math::{make_poly, next_higher_power_of_two, Rem, Zero},
+    test::{
+        bfv_delta, convert_to_polynomial, convert_to_smallint, strip_trailing_value, LatticeProblem,
+    },
+    InnerProductVerifierKnowledge, LogProof, LogProofGenerators, LogProofProverKnowledge,
+    LogProofTranscript,
+};
+
+fn full_knowledge_proof<F, const N: usize>(
```
rickwebiii, line 20: I think we should call this "test_seal_linear_relation"

seal_fhe/src/encryptor_decryptor.rs
rickwebiii: Hmm... the thing we want to indicate is "encrypt and return the data needed to make a ZKP." Maybe "encrypt_and_return_zkp_data"? component is a very generic term.

rickwebiii (approved):

logproof/tests/seal.rs
ryanorendorff: Sounds good! Changed

seal_fhe/src/error.rs
```diff
@@ -51,6 +51,10 @@ pub enum Error {
     #[error("Plain modulus not set")]
     PlainModulusNotSet,
 
+    /// User failed to set a coefficient modulus.
+    #[error("Cannot reduce the modulus from a set size of 1 to 0")]
+    CoefficientModulusSingletonSetDrop,
```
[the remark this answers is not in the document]
ryanorendorff: Changed.

PR #292, opened 2023-08-02
grouping proc macro

matthew-liu801:


examples/debugger/src/main.rs
```diff
@@ -8,19 +9,52 @@ use sunscreen::{
 };
 
 fn main() {
+    #[debug_program]
```
rickwebiii, line 12: I think we should just call this `debug` and users can do `sunscreen::debug` to be clear about what this is.

sunscreen_compiler_common/src/context.rs
```diff
@@ -316,9 +316,9 @@ where
 /**
  * Program groups are given names as strings.
  */
-pub type Group = String;
+// pub type Group = String;
```
rickwebiii, line 319: Remove

sunscreen_compiler_common/src/context.rs
```diff
@@ -341,7 +341,11 @@ where
     /**
      * Represents the program context. Tracks groups of nodes in the compilation graph.
      */
-    pub group_stack: Vec<Group>,
+    pub group_stack: Vec<u64>,
```
rickwebiii, line 344: I expect a stack of `Group`s here. When you call add_node, you pass the entire stack to the group trie and lookup the group_id.

sunscreen_compiler_common/src/context.rs
```diff
@@ -388,20 +394,33 @@ where
...
-
             self.graph
                 .metadata
                 .stack_lookup
                 .id_data_lookup
                 .entry(stack_id)
                 .or_insert(stack_frames);
 
-            self.graph.add_node(NodeInfo {
+            let group_id = *self.group_stack.last().unwrap();
```
rickwebiii: This should be a lookup in the group trie structure thingy.

sunscreen_compiler_common/src/context.rs
```diff
@@ -388,20 +394,33 @@ where
...
+
+            let idx = self.graph.add_node(NodeInfo {
                 operation,
                 group_id,
                 stack_id,
-            })
+            });
+
+            for i in &self.group_stack {
```
rickwebiii, line 412: I thought we were going to dynamically compute the nodes in a group during the REST call?

sunscreen_compiler_common/src/context.rs
```diff
@@ -444,5 +463,48 @@ where
      */
     pub fn add_edge(&mut self, from: NodeIndex, to: NodeIndex, edge: EdgeInfo) {
         self.graph.add_edge(from, to, edge);
+
+        #[cfg(feature = "debugger")]
```
rickwebiii, line 467: Maybe do this lazily in the REST handler?

rickwebiii: I won't die on this hill, but we should avoid storing redundant data when we can avoid it.

sunscreen_compiler_common/src/context.rs
```diff
@@ -444,5 +463,48 @@ where
...
+            }
+        }
+    }
+
+    #[cfg(feature = "debugger")]
+    /**
+     *
+     */
+    pub fn push_group(&mut self, name: String, source: String) {
+        let parent = *self.group_stack.last().unwrap_or(&0);
```
rickwebiii, line 495: I'd make parent an `Option<u64>` on Group.

sunscreen_compiler_common/src/context.rs
```diff
@@ -444,5 +463,48 @@ where
...
+            format!("{}_{}", name, self.next_group_id),
+            parent,
+            source,
+        );
+        self.group_stack.push(self.next_group_id);
+        self.graph
+            .metadata
+            .group_lookup
+            .id_data_lookup
+            .insert(self.next_group_id, group);
```
rickwebiii, line 507: I see, you're using parent to chase up the stack and find the root rather than using the entire group stack as keys... I think that can work.

sunscreen_compiler_common/src/lookup.rs
```diff
@@ -101,7 +101,44 @@ impl Default for StackFrameLookup {
...
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct Group {
+    /** */
+    pub id: u64,
+    /** */
+    pub name: String,
+    /** */
+    pub source: String,
+    /** */
+    pub parent: u64,
```
rickwebiii, line 118: Make this `Option<u64>`, as the root doesn't have a parent. Since the root's node id is also zero, I suspect you're going to have bugs figuring out if you've hit the root or not.

sunscreen_compiler_common/src/context.rs
rickwebiii: Actually, I think we can keep this as is.

sunscreen_compiler_common/src/lookup.rs
```diff
@@ -125,7 +161,6 @@ impl GroupLookup {
     pub fn new() -> Self {
```
rickwebiii, line 161: Add a simple method here

```rust
pub fn children_of(&self, id: u64) -> Vec<u64> {
  self.id_data.lookup.values().iter().map(|x| x.parent == id).collect()
}
```

which you can use to figure out the children of this group.

sunscreen_runtime/src/debugger/server.rs
```diff
@@ -321,3 +322,42 @@ pub async fn get_stack_trace(
     }
     Ok(HttpResponse::NotFound().body(format!("Session {} not found", session)))
 }
+
+#[cfg(feature = "debugger")]
+#[get("sessions/{session}/groups/{groupid}")]
```
rickwebiii, line 327: I think this API should return something like

```rust
enum NodeData {
  Group(u64,) // GroupId
  FheOperation(FheOperation),
  ZkpOperation(ZkpOperation)
}

// Return this thing
StableGraph<NodeData, EdgeInfo>
```

This should be the graph to display for this group.

rickwebiii: You can construct this graph dynamically from the full graph by doing

```rust
let child_groups = id_data_lookup.children_for(groupid).iter().collect::<HashSet<u64>>();

let mut graph = match curr_session {
  Session::BfvSession(curr_session) => {
    curr_session.graph.filter_map(|_, n| {
      if n.group_id == groupid {
        Some(DisplayNodeInfo::FheOperation(n))
      }
    }, |_, e| Some(e))
  },
  Session::ZkpSession(curr_session) => {
    curr_session.graph.filter_map(|_, n| {
      if n.group_id == groupid {
        Some(DisplayNodeInfo::ZkpOperation(n))
      }
    }, |_, e| Some(e))
  }
};

// Map group ids to the node ids you add to the graph.
let child_groups = child_groups.iter().map(|g| (g, graph.add_node(DisplayNodeInfo::Group(g)))).collect::<HashMap<u64, NodeIndex>>();

// Iterate nodes we added and see if we need to hook them up to a group node
for n in graph.node_indices() {
  for e in graph.neighbors_directed(n, Direction::Outgoing) {
    if child_groups.contains_key(graph.node_weight(e.target()).unwrap().group_id) {
      graph.add_edge(n, e.target, DisplayEdgeInfo::GroupEdge);
    }
  }

  for e in graph.neighbors_directed(n, Direction::Incoming) {
    if child_groups.contains_key(graph.node_weight(e.source()).unwrap().group_id) {
      graph.add_edge(n, e.target, DisplayEdgeInfo::GroupEdge);
    }
  }
}
```

rickwebiii: That should replace all your complex data structures.

.vscode/launch.json
```diff
@@ -453,7 +453,8 @@
                     "test",
                     "--no-run",
                     "--lib",
-                    "--package=sunscreen_backend"
+                    "--package=sunscreen_backend",
+                    "--features=debugger"
```
rickwebiii, line 457: Remove the debugger feature

examples/sudoku_zkp/Cargo.toml
```diff
@@ -4,5 +4,5 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-sunscreen = { path = "../../sunscreen" }
-sunscreen_zkp_backend = { path = "../../sunscreen_zkp_backend" }
\ No newline at end of file
+sunscreen = { path = "../../sunscreen", features = ["debugger"]}
```
rickwebiii, line 7: Remove this feature. I think the debugger example should be the only one to use this.

examples/sudoku_zkp/src/main.rs
```diff
@@ -49,33 +51,52 @@ fn main() {
     let verify = runtime.verify(prog, &proof, cons, vec![]);
 
     assert!(verify.is_ok());
+
+    loop {
```
rickwebiii, line 55: Don't put this in your zkp example.

examples/zkp_debugger/src/main.rs
```diff
@@ -1,14 +1,17 @@
...
+};
 use sunscreen_zkp_backend::{bulletproofs::BulletproofsBackend, BackendField, ZkpBackend};
 
 fn main() {
     type BPField = NativeField<<BulletproofsBackend as ZkpBackend>::Field>;
 
     #[zkp_program(backend = "bulletproofs")]
     fn prove_sum_eq<F: BackendField>(a: NativeField<F>, b: NativeField<F>, c: NativeField<F>) {
-        (a + b).constrain_eq(c); // not satisfied
+        (a + b).constrain_lt_bounded(c, 8); // not satisfied
```
rickwebiii, line 14: ?

sunscreen_backend/src/transforms/insert_relinearizations.rs
```diff
@@ -64,13 +64,50 @@ pub fn apply_insert_relinearizations(ir: &mut FheProgram) {
...
+    {
+        let mut group_updates: Vec<(u64, u64)> = vec![];
+        for i in ir.graph.node_indices() {
+            group_updates.push((
+                ir.graph.node_weight(i).unwrap().group_id,
+                i.index().try_into().unwrap(),
+            ));
+        }
+        for (g, i) in group_updates {
+            group_insert_recursive(&g, i, ir);
```
rickwebiii, line 78: I had made some comments on your other review to lazily compute the display graph when the user makes a REST call so you don't have a bunch of extra data structures floating around that you have to keep up to date.

I still strongly think we should do that and remove this code. The only thing this compiler transform should do WRT to groups is ensure the relin node has the same group id as the node that comes before it.

sunscreen_runtime/debugger-frontend/src/App.tsx
```diff
@@ -306,24 +384,27 @@ const App = () => {
...
-    }
-    update()
+    setGroupStack([0]);
   }, [session])
 
   useEffect(() => {
-      setProblematicNodes(currGraph)
-    }, [currGraph]
+      async function update() {
+        const newGraph = groupToGraph(await fetch(`/sessions/${session}/groups/${groupStack.at(-1)}`).then(j => j.json()))
```
rickwebiii, line 401: I'd prefer not to mix promises (i.e. .then) when async await. Do something like 
```ts
groupToGraph(async () => {
  const response = await fetch(`/sessions/${session}/groups/${groupStack.at(-1)}`);
  return await response.json();
});
```

sunscreen_runtime/src/debugger/server.rs
```diff
@@ -321,3 +358,291 @@ pub async fn get_stack_trace(
...
+
+    if let Some(curr_session) = sessions.get(&session) {
+        match curr_session {
+            Session::BfvSession(s) => {
+                let lookup = &s.graph.metadata.group_lookup;
+
+                if let Some(curr_group) = lookup.id_data_lookup.get(&groupid.try_into().unwrap()) {
+                    let child_groups = lookup.children_of(groupid.try_into().unwrap());
+
+                    let mut graph = s.graph.filter_map(
```
rickwebiii, line 378: I would move this to a helper function called `generate_fhe_display_graph`. Ditto for the ZKP case below and share as much code as you can.

sunscreen_backend/src/transforms/insert_relinearizations.rs
rickwebiii: Actually, you should be able to just remove this code?

sunscreen_runtime/src/metadata.rs
```diff
@@ -198,7 +198,7 @@ pub struct FheProgramMetadata {
     pub name: String,
 }
 
-#[derive(Clone, Serialize, Deserialize)]
+#[derive(Clone, Serialize)]
```
rickwebiii, line 201: Can we not keep this `Deserialize`?

rickwebiii (approved):

sunscreen_runtime/src/metadata.rs
bryanyli: I don't think Operation implements Deserialize

PR #297, opened 2023-08-07
Add restore-keys to github actions

ryanorendorff:


ryanorendorff:
My understanding is that if you change the lock file (and therefore get a cache miss), the restore key will restore the last thing it saw in cache anyway. This is attempting to closer match the local development experience where when you update the lock file it doesn't completely start from scratch but from the last build you had.

PR #303, opened 2023-08-18
Add deterministic encryption

ryanorendorff:
Adds a deterministic encryption function that is behind the `deterministic` feature flag. This can be accessed from the added `encrypt_deterministic` and `encrypt_return_components_deterministic` functions. Normal encryption works as it did before. 

If we want to merge this then [SEAL deterministic](https://github.com/Sunscreen-tech/SEAL/tree/deterministic) branch should be merged. If we want to keep this as a separate branch, then the SEAL branch can stay as it is.


seal_fhe/src/encryptor_decryptor.rs
```diff
@@ -469,4 +566,58 @@ mod tests {
 
         assert_eq!(data, data_2);
     }
+
+    #[cfg(feature = "deterministic")]
```
samtay, line 570: If you want to hold off on the changes to CI, we could have easier test coverage of this by just always defining these functions and always testing them, but only exporting them when given the feature flag. Something like
```rust
mod deterministic {
     // functions here
}
#[cfg(feature = "deterministic")]
pub use deterministic::*;

samtay (approved):
lgtm

PR #308, opened 2023-08-24
Add deterministic linux hash, gh actions workflow

ryanorendorff:


samtay (approved):
Yikes we don't even test the bulletproofs flag currently :grimacing: but out of scope for this PR

PR #313, opened 2023-09-09
Rweber/zq

rickwebiii:
seal test is still failing, but everything else passes including benchmarks.


logproof/src/rings.rs
```diff
@@ -0,0 +1,349 @@
...
+    fn mod_switch_unsigned(&self) -> Zq<N, Q2> {
+        assert!(N >= M);
+
+        Zq::try_from(extend_bigint(&self.into_bigint())).unwrap()
+    }
+
+    fn mod_switch_signed(&self) -> Zq<N, Q2> {
+        let val = self.into_bigint();
+
+        let mod_div_2 = Q1::MODULUS >> 1;
```
rickwebiii, line 146: This is already available as a constant.

sunscreen_math/src/poly/mod.rs
```diff
@@ -1,32 +1,179 @@
...
+
+        let iter_count = lhs_degree - rhs_degree + 1;
+        let mut q = Polynomial {
+            coeffs: vec![R::zero(); iter_count],
+        };
+
+        for i in 0..iter_count {
+            // Normally, we would compute the scale factor as coeff_i(rem) * coeff_i(rhs)^-1,
+            // but inverse isn't defined for rings. Since we leverage the fact that the
+            // leading coefficient
```
samtay, line 141: looks like this comment got cut short?

sunscreen_math/src/poly/mod.rs
```diff
@@ -1,32 +1,179 @@
...
+    /// In order for polynomial division to work in a ring, `rhs` has restrictions.
+    /// Specifically, the highest order non-zero coefficient must be 1 so as to avoid
+    /// inverse operations. While multiplicative inverses are not guaranteed to exist
+    /// for a [`Ring`] element, the inverse of one will always be one.
+    ///
+    /// # Panics
+    /// If the divisor's leading non-zero coefficient isn't one.
+    ///
+    /// If rhs is the zero polynomial.
+    pub fn vartime_div_rem_restricted_rhs(&self, rhs: &Self) -> (Self, Self) {
```
samtay, line 116: maybe want to add a test for this function - unless it's covered elsewhere and I missed it

sunscreen_math/src/poly/mod.rs
```diff
@@ -75,15 +252,75 @@ where
...
+    T: Ring,
+{
+    type Output = Polynomial<T>;
+
+    fn mul(self, rhs: &T) -> Self::Output {
+        Self::Output {
+            coeffs: self
+                .coeffs
+                .iter()
+                .map(|x| x.clone() * rhs)
```
samtay, line 267: Huh... this doesn't look right to me :thinking:  Why component wise coefficient multiplication here?

samtay: I'm guessing this is made possible by the Barret backend. If so, we should perhaps document that correctness depends on choices of `T`, or better yet have a trait constraint that enforces it, as I don't think every ring would work here. But I am brain dead so maybe I'm missing something.

samtay (approved):

sunscreen_math/src/poly/mod.rs
samtay: Oh my bad - I see, this isn't component-wise

sunscreen_math/src/poly/mod.rs
samtay: nvm - this is tested below

rickwebiii: Check the end of the file.

sunscreen_math/src/poly/mod.rs
rickwebiii: This has nothing to do with the Barrett backend and the `Ring` doesn't imply Barrett reductions. Coefficient-wise multiplication is sensible on polynomials in NTT or FFT form, but that's completely unrelated. Look again and you'll see we're multiplying each coefficient in lhs by `rhs`, not component-wise multiplication. This is simple scalar multiplication.

Take for example, `Z[X] x Z`. `a in Z` is equivalent to the polynomial `a*x^0`, so it's well defined.

PR #323, opened 2023-10-14
Fix ToUInt: 2^n doesn't fit n bits

JoseSK999:
Also a small change to the AssertBinary comment to reflect the actual constraint used

### DIFF
--- a/sunscreen/src/types/zkp/gadgets/binary.rs
+++ b/sunscreen/src/types/zkp/gadgets/binary.rs
@@ -33,7 +33,7 @@ impl Gadget for ToUInt {
             return Err(ZkpError::gadget_error("Cannot create 0-bit uint."));
         }
 
-        if *val > BigInt::ONE.shl_vartime(self.n) {
+        if *val > BigInt::ONE.shl_vartime(self.n).wrapping_sub(&BigInt::ONE) {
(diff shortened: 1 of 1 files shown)


sunscreen/src/types/zkp/gadgets/binary.rs
```diff
@@ -33,7 +33,7 @@ impl Gadget for ToUInt {
             return Err(ZkpError::gadget_error("Cannot create 0-bit uint."));
         }
 
-        if *val > BigInt::ONE.shl_vartime(self.n) {
+        if *val >= BigInt::ONE.shl_vartime(self.n) {
```
samtay, line 36: Good catch! This is actually still incorrect when `n == 512`, however, as `BigInt::ONE.shl_vartime(512) == BigInt::ZERO`. I believe the correct fix is to change this line to

```suggestion
        if *val > BigInt::ONE.shl_vartime(self.n).wrapping_sub(&BigInt::ONE) {
```

samtay (changes requested):
Thanks for your contribution! Just one small change requested.

JoseSK999:
Right @samtay, fixed

samtay (approved):

PR #328, opened 2023-10-19
Linked prove and verify functions

ryanorendorff:
Add the `LinkedProof` struct that facilitates making a linked proof between SDLP and an R1CS BP.

A few open questions (hence why this is starting as a draft):

1. Is this the low level API we want? By passing in a `LatticeProblem` we can support any variation on the SDLP we want, so that seems nice. 
2. Should the `create` function take in a `LatticeProblem` or a `logproof::ProverKnowledge`? They are equivalent, one just has more layers. I'm leaning toward removing `LatticeProblem` completely and only using `ProverKnowledge`.
3. I don't quite understand why the test does not compile with a `error: the trait bound ``sunscreen::types::zkp::Field<BulletproofsFieldSpec>: ToNativeFields`` is not satisfied` when the exact same code copied into an independent program does find the `ToNativeFields` implementation.
4. Either compilation for WASM needs to be fixed when including the `logproof` crate or the linked proofs should be behind a feature flag.
5. The documentation is a draft. We can either keep it where it is or move it into docs and import it in this file.

### DIFF
--- a/logproof/src/test.rs
+++ b/logproof/src/test.rs
@@ -402,12 +402,16 @@ where
     // Calculate the bounds for each element in S. m and r should have bounds of
     // the plaintext modulus, while u is ternary and e is sampled from the
     // centered binomial distribution with a standard deviation of 3.2.
-    let m_bounds = Bounds(
-        m_coeffs
-            .iter()
-            .map(|c| if *c == 0 { 0 } else { plain_modulus.value() })
-            .collect::<Vec<u64>>(),
-    );
+
+    let m_bounds = if batch_encoder {
+        Bounds(vec![plain_modulus.value(); m_coeffs.len()])
+    } else {
+        let mut bounds = vec![0; m_coeffs.len()];
+        bounds[0] = plain_modulus.value();
+
+        Bounds(bounds)
+    };
+
     let r_bounds = Bounds(vec![plain_modulus.value(); r_coeffs.len()]);
     let u_bounds = Bounds(vec![2; u_small[0].len()]);
     let e1_bounds = Bounds(vec![16; e_small[0].len()]);
--- a/logproof/src/test.rs
+++ b/logproof/src/test.rs
@@ -3,10 +3,14 @@
  * production, only for testing.
  */
 use crypto_bigint::{NonZero, Uint};
-use seal_fhe::{Modulus, PolynomialArray};
+use seal_fhe::{
+    BFVEncoder, BFVScalarEncoder, BfvEncryptionParametersBuilder, CoefficientModulus, Context,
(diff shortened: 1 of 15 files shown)


logproof/src/test.rs
```diff
@@ -176,3 +180,241 @@ pub fn bfv_delta<const N: usize>(coeff_modulus: ZqRistretto, plaintext_modulus:
...
 }
+
+/**
+ * Generate a lattice problem for the BFV scheme.
+ */
+pub fn seal_bfv_encryption_linear_relation<B, const N: usize>(
+    message: u64,
+    degree: u64,
+    plain_modulus: u64,
+    batch_encoder: bool,
```
rickwebiii, line 191: This isn't quite the shape I would expect, but not sure it matters much since we'll replace it anyways.

sunscreen_runtime/src/linked.rs
```diff
@@ -0,0 +1,611 @@
...
+     * let verified = linked_verify(
+     *     &lp,
+     *     valid_transaction_zkp,
+     *     vec![BulletproofsField::from(balance)], // Public inputs
+     *     vec![],                                 // Constant inputs
+     * );
+     * println!("Verified linked proof: {}", verified);
+     * ```
+     */
+    pub fn create<I>(
```
rickwebiii, line 326: We'll need to expose a function for the prover and verifier independently, since the verifier won't have the witness.

rickwebiii: But generally I think this is almost right. It should take a prover or verifier knowledge and link the proofs.

rickwebiii (approved):

logproof/src/test.rs
ryanorendorff: It is just a test function I pulled out to use in both the linked proof and the SDLP tests. It's just a test function, not a general API.

sunscreen_runtime/src/linked.rs
```diff
@@ -0,0 +1,343 @@
...
+        shared_indices: &[(usize, usize)],
+        program: &CompiledZkpProgram,
+        private_inputs: Vec<I>,
+        public_inputs: Vec<I>,
+        constant_inputs: Vec<I>,
+    ) -> Result<Self, sunscreen_zkp_backend::Error>
+    where
+        I: Into<ZkpProgramInput> + Clone,
+    {
+        let backend = BulletproofsBackend::new();
```
samtay, line 152: These changes leak the bulletproof backend implementation without the feature flag enabled - not sure what we want to do about that.

ryanorendorff: I am going to put this behind a feature flag anyway (or potentially just the bulletproofs feature flag) because otherwise the wasm build won't work

PR #337, opened 2024-01-16
Generating SDLP for BFV statements

samtay:
This PR contains the SDLP prover/verifier knowledge generation from a set of BFV encryption statements. Everything here has been added to `logproof` and written in terms of `seal_fhe`, so it is a bit of a lower level API than what will be used by a sunscreen library consumer.

Next up will be a higher level API added to `sunscreen_runtime`, making use of the new `logproof::bfv_statement` under the hood. Once I've adapted everything in that crate to use the new swag, I can also remove the existing `logproof::test`, `LatticeProblem`, etc.

### DIFF
--- a/examples/calculator_fractional/src/main.rs
+++ b/examples/calculator_fractional/src/main.rs
@@ -121,7 +121,7 @@ fn encrypt_term(runtime: &FheRuntime, public_key: &PublicKey, input: Term) -> Te
         Term::Ans => Term::Ans,
         Term::F64(v) => Term::Encrypted(
             runtime
-                .encrypt(Fractional::<64>::try_from(v).unwrap(), public_key)
+                .encrypt(Fractional::<64>::from(v), public_key)
                 .unwrap(),
         ),
         _ => {
@@ -250,7 +250,7 @@ fn bob(
         let runtime = FheRuntime::new(app.params()).unwrap();
 
         let mut ans = runtime
-            .encrypt(Fractional::<64>::try_from(0f64).unwrap(), &public_key)
+            .encrypt(Fractional::<64>::from(0f64), &public_key)
             .unwrap();
 
         loop {
--- a/logproof/src/bfv_statement.rs
+++ b/logproof/src/bfv_statement.rs
@@ -20,12 +20,10 @@ use crate::{
 };
 
 /// In SEAL, `u` is sampled from a ternary distribution.
-// TODO if bounds are in u64 then perhaps this should be 3?
 const U_COEFFICIENT_BOUND: u64 = 2;
 /// In SEAL, `e` is sampled from a centered binomial distribution with std dev 3.2.
 const E_COEFFICIENT_BOUND: u64 = 16;
 /// In SEAL, secret keys are sampled from a ternary distribution.
-// TODO if bounds are in u64 then perhaps this should be 3?
 const S_COEFFICIENT_BOUND: u64 = 2;
 
 /// A proof statement verifying that a ciphertext is an encryption of a known plaintext message.
--- a/logproof/src/bfv_statement.rs
+++ b/logproof/src/bfv_statement.rs
@@ -34,37 +34,50 @@ const S_COEFFICIENT_BOUND: u64 = 2;
 /// hence multiple proof statements.
 #[derive(Debug)]
 pub enum BfvProofStatement<C, P> {
+    /// A statement that the ciphertext symmetrically encrypts the identified message.
     PrivateKeyEncryption {
+        /// Column index in the A matrix, or equivalently the index of the message slice provided
+        /// when generating the prover knowledge.
         message_id: usize,
+        /// The ciphertext of the encryption statement.
         ciphertext: C,
     },
+    /// A statement that the ciphertext asymmetrically encrypts the identified message.
     PublicKeyEncryption {
+        /// Column index in the A matrix, or equivalently the index of the message slice provided
+        /// when generating the prover knowledge.
         message_id: usize,
+        /// The ciphertext of the encryption statement.
         ciphertext: C,
+        /// The public key of the encryption statement.
         public_key: P,
     },
 }
 
 impl<C, P> BfvProofStatement<C, P> {
-    fn message_id(&self) -> usize {
+    /// Get the message index of this statement.
+    pub fn message_id(&self) -> usize {
         match self {
             BfvProofStatement::PrivateKeyEncryption { message_id, .. } => *message_id,
             BfvProofStatement::PublicKeyEncryption { message_id, .. } => *message_id,
         }
     }
 
-    fn ciphertext(&self) -> &C {
+    /// Get the ciphertext of this statement.
+    pub fn ciphertext(&self) -> &C {
         match self {
             BfvProofStatement::PrivateKeyEncryption { ciphertext, .. } => ciphertext,
             BfvProofStatement::PublicKeyEncryption { ciphertext, .. } => ciphertext,
         }
     }
 
-    fn is_public(&self) -> bool {
+    /// Return whether or not this is a public encryption statement.
+    pub fn is_public(&self) -> bool {
         matches!(self, BfvProofStatement::PublicKeyEncryption { .. })
     }
 
-    fn is_private(&self) -> bool {
+    /// Return whether or not this is a private encryption statement.
+    pub fn is_private(&self) -> bool {
         matches!(self, BfvProofStatement::PrivateKeyEncryption { .. })
     }
 }
@@ -72,6 +85,7 @@ impl<C, P> BfvProofStatement<C, P> {
 /// A witness for a [`BfvProofStatement`].
 #[derive(Debug)]
 pub enum BfvWitness<S> {
+    /// A witness for the [`BfvProofStatement::PrivateKeyEncryption`] variant.
     PrivateKeyEncryption {
         /// The private key used for the encryption.
         private_key: S,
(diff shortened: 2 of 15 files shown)


logproof/src/bfv_statement.rs
```diff
@@ -0,0 +1,863 @@
...
+
+use crate::{
+    linear_algebra::{Matrix, PolynomialMatrix},
+    rings::ZqRistretto,
+    Bounds, LogProofProverKnowledge, LogProofVerifierKnowledge,
+};
+
+/// In SEAL, `u` is sampled from a ternary distribution.
+// TODO if bounds are in u64 then perhaps this should be 3?
+const U_COEFFICIENT_BOUND: u64 = 2;
```
samtay, line 24: Calling this out for those more familiar... seems like it should be 3? But existing bounds were using 2 on the ternary polynomials.

ryanorendorff: The 2 just refers to the max absolute value of the bound. So since the value can be in {-1, 0, 1}, the bound should be 2. I believe, correct me if I am remembering incorrectly.

samtay: I think you are right, after looking [here](https://github.com/Sunscreen-tech/Sunscreen/blob/5cccd88ac0ee4ab53287a4ee0a0fed2126abd2b4/logproof/src/linear_relation.rs#L1138). I've removed the TODO comments.

Also the tests would have undoubtedly failed by now if the bound 2 was too low, so there's that.

examples/calculator_fractional/src/main.rs
```diff
@@ -121,7 +121,7 @@ fn encrypt_term(runtime: &FheRuntime, public_key: &PublicKey, input: Term) -> Te
         Term::Ans => Term::Ans,
         Term::F64(v) => Term::Encrypted(
             runtime
-                .encrypt(Fractional::<64>::try_from(v).unwrap(), public_key)
+                .encrypt(Fractional::<64>::from(v), public_key)
```
rickwebiii, line 124: Why is this allowed? What if the value to pass for conversion is NaN, Infinity out of range?

logproof/src/bfv_statement.rs
```diff
@@ -0,0 +1,861 @@
...
+use crate::{
+    linear_algebra::{Matrix, PolynomialMatrix},
+    rings::ZqRistretto,
+    Bounds, LogProofProverKnowledge, LogProofVerifierKnowledge,
+};
+
+/// In SEAL, `u` is sampled from a ternary distribution.
+const U_COEFFICIENT_BOUND: u64 = 2;
+/// In SEAL, `e` is sampled from a centered binomial distribution with std dev 3.2.
+const E_COEFFICIENT_BOUND: u64 = 16;
```
rickwebiii, line 25: Seal caps at 6 standard deviations, which is 18.6 IIRC. So, raise this to 19.

logproof/src/bfv_statement.rs
```diff
@@ -0,0 +1,861 @@
...
+    B: BarrettConfig<N>,
+    C: AsRef<Ciphertext>,
+    P: AsRef<PublicKey>,
+    T: StatementParams,
+{
+    let mut offsets = IdxOffsets::new(statements);
+    let (rows, cols) = offsets.a_shape();
+    let mut a = PolynomialMatrix::new(rows, cols);
+
+    let d = Polynomial {
```
rickwebiii, line 252: Is this proving textbook BFV, or SEAL's variant that Ryan got working?

logproof/src/bfv_statement.rs
```diff
@@ -0,0 +1,861 @@
...
+                    .collect::<Vec<_>>(),
+                Zq::zero(),
+            ),
+        }
+    }
+}
+
+impl<const N: usize, B: BarrettConfig<N>> AsPolynomial<Z<N, B>> for SecretKey {
+    fn as_poly(&self) -> Polynomial<Z<N, B>> {
+        todo!()
```
rickwebiii, line 516: Delete this impl or impl it.

logproof/src/bfv_statement.rs
```diff
@@ -0,0 +1,861 @@
...
+                .map(|_| self.random_plaintext())
+                .collect::<Vec<_>>();
+
+            let mut statements = Vec::with_capacity(num_statements);
+            let mut witness = Vec::with_capacity(num_statements);
+
+            // statements without duplicate messages
+            for (i, msg) in messages.iter().enumerate() {
+                let (ct, u, e, r) = self.encryptor.encrypt_return_components(msg).unwrap();
+                statements.push(BfvProofStatement::PublicKeyEncryption {
```
rickwebiii, line 815: Do we have any tests with secret encryptions?

logproof/src/linear_relation.rs
```diff
@@ -1501,7 +1486,7 @@ mod test {
         let mut vals = vec![];
 
         for i in 0..257 {
-            let ring_val = Zr::try_from(i).unwrap();
+            let ring_val = Zr::from(i);
```
rickwebiii, line 1489: I forget if I provided a from implementation. What does it do if i > p?

seal_fhe/src/key_generator.rs
```diff
@@ -418,6 +425,20 @@ impl Clone for SecretKey {
     }
 }
 
+impl core::fmt::Debug for SecretKey {
```
rickwebiii, line 428: I guess this isn't the worst thing, but I have seen security issues at previous companies where people accidentally log credentials or secrets. Is it better to impl Debug and print **** as you've done or to not impl it at all?

sunscreen/src/types/bfv/batched.rs
```diff
@@ -676,52 +676,52 @@ mod tests {
...
-        let a = Batched::<4>::try_from(A_VEC).unwrap();
+        let a = Batched::<4>::from(A_VEC);
 
         assert_eq!(-a, [[-1, -2, -3, -4], [-5, -6, -7, -8]].into());
     }
 
     #[test]
     fn can_shl_non_fhe() {
-        let a = Batched::<4>::try_from(A_VEC).unwrap();
+        let a = Batched::<4>::from(A_VEC);
```
rickwebiii, line 710: What happens on out of range?

sunscreen_curve25519
rickwebiii, line 1: Is there a reason to do this? I wouldn't remove imports from forks even if unused unless there's a really compelling reason.

rickwebiii (approved):

examples/calculator_fractional/src/main.rs
samtay: All of these were just clippy fixes. The warning is basically telling you when there's a `From` instance, just use it, rather than introduce the noise of trying something that is infallible. The existing code works because `T` impls `TryFrom<U>` where `T: From<U>`, with an infallible error type.

So, this code is semantically the same, just less noise.

> What if the value to pass for conversion is NaN, Infinity out of range?

If we want to handle these, then we should implement `TryFrom<f64>` rather than [`From<f64>` for `Fractional`](https://github.com/Sunscreen-tech/Sunscreen/blob/5cccd88ac0ee4ab53287a4ee0a0fed2126abd2b4/sunscreen/src/types/bfv/fractional.rs#L560-L564).

logproof/src/bfv_statement.rs
samtay: Ryan's - much of the `seal_fhe -> sunscreen_math` conversions are copied from e.g. `logproof::test`. (Which can be removed after we adapt `sunscreen_runtime` to use `logproof::bfv_statement`).

logproof/src/bfv_statement.rs
samtay: I'd prefer to leave it for now.. it's the only piece that needs to happen after we expose symmetric encryption from SEAL. If I remove the impl, I either need to comment out all of the private key encryption functionality in this module or do something nonsensical, like use a dummy polynomial, and I'd rather leave an explicit panic than something blatantly wrong. Would you like `unimplemented!()` better?

logproof/src/bfv_statement.rs
samtay: No.. as discussed, symmetric encryptions are on hold until after they're exposed from SEAL. Hence the `todo!()` above.

logproof/src/linear_relation.rs
samtay: It [doesn't seem to check that](https://github.com/Sunscreen-tech/Sunscreen/blob/5cccd88ac0ee4ab53287a4ee0a0fed2126abd2b4/sunscreen_math/src/ring/mod.rs#L533-L537).

Also just to be clear, these `try_from -> from` warnings are _because_ there are `From` implementations and _not_ `TryFrom` implementations (offering both would conflict).

seal_fhe/src/key_generator.rs
samtay: My personal opinion is that it is better to do this. For one, you all but eliminate the chances that someone will accidentally add a `#[derive(Debug)]`, which is the most likely way that someone will accidentally log the secret. Secondly, for convenience, since things like tests, assertions, etc. rely on a `Debug` impl.

But in this case it really doesn't matter, because the derived impl would just print a pointer value :upside_down_face:. So I did this for convenience, as there is no negative as far as I can see.

sunscreen_curve25519
samtay: Again this is just a clippy warning. I verified that removing it has no effect. I thought perhaps the things imported by `pub use *` were being exposed higher up in the mod structure, but they were not.

samtay: I can alternatively just add a `#[allow(clippy::_)]` if you'd prefer?

samtay: I'll open a follow up PR today to remove our clippy fixes from the fork and keep the diff more minimal.

PR #339, opened 2024-02-01
Proof linking

samtay:
### summary
- New [example](https://github.com/Sunscreen-tech/Sunscreen/blob/samtay/proof-linking/examples/private_tx_linkedproof/src/main.rs) of linked proof API usage.
- Some simplifications to the types in [`logproof::bfv_statement`](https://github.com/Sunscreen-tech/Sunscreen/pull/339/files#diff-30637456809a880b2cf4e4728cceaf0fd6579ffb9e355438f2eb07c265893245) but also added ability to specify bounds per message, rather than defaulting to plain modulus for the entirety of the lattice dimension.
- Fixed up some of the `Error` types
- Two things that need review in [`logproof::linear_relation`](https://github.com/Sunscreen-tech/Sunscreen/pull/339/files#diff-5678755c608f5bd04e7ab02ebe4b5d9f39368a90a186a3f8225dd80f5dfd60af) (see comments below)
- Removed the now duplicate code in [`logproof::test`](https://github.com/Sunscreen-tech/Sunscreen/pull/339/files#diff-1443004f2fce2e7e091f69951124a5473d9bb9aece75448247cecf66f95a87c3) but left in place the `seal_bfv_encryption_linear_relation` as a test helper method.
- Fixed docs.rs latex inclusion. There unfortunately isn't a great way to have this "just work" both on docs.rs and locally. If you want latex locally just make sure you run with `RUSTDOCFLAGS="--html-in-header sunscreen/docs/assets/katex-header.html"`, relative to the workspace. You also _must_ pass `--no-deps`. For that reason I've chosen not to add a `.config/cargo.toml` as it would break the normal `cargo doc`.
- Removed the math-heavy linked proofs docs, these felt out of place in the API docs, especially now that we have a simpler API on top. Will include them as background in the user docs.
- Added a [`ZkpProgramFn::Share`](https://github.com/Sunscreen-tech/Sunscreen/pull/339/files#diff-d39b0282f29626cc16670dc71e1017349ee9143fe7d4eb741ff34db9f7e2c9f2) associated type indicator that gets set in the presence of new [`#[shared]`](https://github.com/Sunscreen-tech/Sunscreen/pull/339/files#diff-66e1b2814215e3dc3d0bffffad57c486129788b0f9c02aef0bbadb4e76961d16) arguments, which must implement `CreateSharedZkpProgramInput` rather than `CreateZkpProgramInput`. Such inputs need a plaintext modulus to be created, hence `ZkpProgramFn::build` now takes a `Share::Input` which is a `u64` for type `Shared`.
- Therefore the [`Compiler`](https://github.com/Sunscreen-tech/Sunscreen/pull/339/files#diff-83d42a6ac306ef763c561ed4cb66cd07d6069548cec5067171f35211e83cb35b) now threads this through if the ZKP program fn is of the shared variety.
- This plaintext modulus then gets used to determine the number of inputs necessary for the ZKP type via the [`DynamicNumFieldElements`](https://github.com/Sunscreen-tech/Sunscreen/pull/339/files#diff-b723be9ecd5ddccecebb916c04c728d657bd80eea967167560ab638c06726929) trait.
- Once in the circuit, the [`AsFieldElement`](https://github.com/Sunscreen-tech/Sunscreen/pull/339/files#diff-73b1ae699e8bfa7abb16a27ffeaddcd298f084076877cf4d03c151bf536a384f) trait handles the actual decoding of those 2s complement encoded coefficients into a `Field` equivalent value.
- I've moved the existing proof builders into [`sunscreen_runtime::builder`](https://github.com/Sunscreen-tech/Sunscreen/pull/339/files#diff-21e54445aabeae8ae4b10a0cf0f2e9deb2e54229bc349aca67f8e342a3a5b598) along with the new `LogProofBuilder`. Most of that builder can be intuited from the example linked above. Just a couple things to note:
  - The FHE types that you can share are restricted via the `ShareWithZkp` trait, and in fact each one _determines_ the associated ZKP type. (So you'll get an arg mismatch error if you supply a `.shared_input(ct: Cipher<Signed>)` when the ZKP prog has an arg type of `BfvRational<F>`.
  - Currently there is only a single notion of "sharing" used to cover both cases of sharing with a ZKP program and proving multiple ciphertexts encrypt the same "shared" message within SDLP. I think we need to separate these notions, just wasn't sure how I wanted to do it.
- That reminds me, when wrapping the compiled ZKP program with metadata, I chose to include the call signature like we do for FHE programs, and actually check it at runtime. This maybe wasn't necessary before, but is now a welcome improvement. If you got your arguments wrong, you would see inscrutable errors like "expected 131313 inputs, got 131322" or simply unsatisfiable constraints.
- Sorry but I also have a bunch of noise fixing our doc comments where we chose to use `ignore` instead of the more appropriate `text` attribute on code blocks.

### next
- Somewhere along the way I started this "share" notion. Perhaps I should rename all instances of this to "link"? Thoughts?
- Impl `ShareWithZkp` for more FHE types.
- Implement symmetric encryptions and expose their proof statements in the builder.

### possibly next
- Allow more than just powers of 2 plaintext moduli. I know of two ways to accomplish this:
  1. Have the prover stick the plaintext modulus into the circuit. This is _surprisingly gory_. There's a lot of hairy code that relies on the shared inputs being on just "one end" of the inputs. I thought if I just passed around the appropriate index ranges for the bulletproofs to use, this wouldn't be so bad, but no such luck. I think there's a block of multipliers that is not one-to-one with the shared generators? Not really sure.
  2. Much simpler: stick the plaintext modulus onto the program node as an optional field. This is the approach we took with FHE variables, and is sort of a general way to move compile time logic into runtime logic. It just feels wrong, and like a slippery slope to bloated program nodes.


logproof/src/linear_relation.rs
```diff
@@ -1197,7 +1198,7 @@ impl LogProof {
         // Make sure we have an equal number of values and bounds to serialize
         assert_eq!(values.len(), log_b.len());
 
-        let mut bitvec = BitVec::with_capacity(values.len() * log_b.iter().sum::<u64>() as usize);
+        let mut bitvec = BitVec::with_capacity(log_b.iter().sum::<u64>() as usize);
```
samtay, line 1201: Pretty sure this is correct, but double check me. We zip these two things together, and the `ith` elem in `values` ends up being `log_b[i]` long. Multiplying by the length was probably blowing this allocation up quite a bit.

logproof/src/linear_relation.rs
```diff
@@ -1120,6 +1120,7 @@ impl LogProof {
      * This modifies bitvec in place.
      *
      */
+    // TODO seems incorrect for log_b == 1; {0, 1} -> 0 and {-1} -> 1
     fn to_2s_complement_single<B, const N: usize>(value: &Zq<N, B>, log_b: u64, bitvec: &mut BitVec)
```
samtay, line 1124: Looking at this more closely, it looks like these `log_b` values get computed by
```rust
|b| if *b > 0 { Log2::ceil_log2(b) + 1 } else { 0 }
```
So 1 should never happen. Perhaps I should add a debug assertion that `log_b != 1`?

sunscreen/src/types/zkp/bfv_plaintext.rs
```diff
@@ -0,0 +1,271 @@
...
+    fn to_native_fields(&self) -> Vec<BigInt> {
+        self.0.to_native_fields()
+    }
+}
+
+/// Decode the underlying plaintext polynomial into the field.
+pub trait AsFieldElement<F: FieldSpec> {
+    /**
+     * Return a structure scaled by `x`.
+     */
```
ryanorendorff, line 61: Is this comment correct? Or is it something that converts the field elements into a signed field element?

sunscreen/src/types/zkp/bfv_plaintext.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    #[test]
+    fn can_decode_signed_for_powers_of_two_moduli() {
+        test_plain_modulus(1024);
+        test_plain_modulus(4096);
+        test_plain_modulus(262_144); // default in compiler.rs
+    }
+
+    #[test]
+    fn can_decode_signed_with_coefficients_near_cutoff() {
```
ryanorendorff, line 157: Neat test!

sunscreen/src/types/zkp/program_node.rs
```diff
@@ -163,6 +160,22 @@ where
     }
 }
 
+impl<T> CreateSharedZkpProgramInput for ProgramNode<T>
+where
+    T: SharedZkpType,
+{
+    fn shared_input(plaintext_modulus: u64) -> Self {
```
ryanorendorff, line 167: Should we add a note that this had to be in a certain order? I guess I am a tad concerned that this is called before/after adding the private inputs and then the `ctx.add_private_input` line adds shared input in the wrong order.

sunscreen/tests/linked.rs
ryanorendorff, line 1: Should we add a test where there is more than one shared input?

sunscreen_runtime/src/builder.rs
```diff
@@ -0,0 +1,648 @@
...
+            self.program,
+            self.private_inputs,
+            self.public_inputs,
+            self.constant_inputs,
+        )
+    }
+
+    /// Generate a proof with parameters for that proof system; see
+    /// [`runtime.prove_with_parameters()`][GenericRuntime::prove_with_parameters].
+    pub fn prove_with_parameters(
```
ryanorendorff, line 118: Will people need to use this publically? If the moving around of shared generators and blinding factors is handled then they probably don't need this function exposed on the top level.

sunscreen_runtime/src/builder.rs
```diff
@@ -0,0 +1,648 @@
...
+        /// The number of nonzero coefficients to share between the SDLP and ZKP.
+        ///
+        /// Note that when plaintexts polynomials are shared with ZKP programs, the ZKP program
+        /// assumes the coefficients form a 2s complement encoding. Without this bound, all
+        /// coefficients (up to the lattice degree `N`) are shared and thus the ZKP circuit has to
+        /// compute `2^N`. Of course, each ZKP field element is limited by its underlying
+        /// representation (e.g. a 256-bit scalar for Bulletproofs). Thus, any encoding should keep
+        /// this limit this in mind.
+        ///
+        /// This number should be less than 256.
```
ryanorendorff, line 269: Why less than 256? What if they want to share an entire BFV 1024 plaintext?

sunscreen_runtime/src/builder.rs
```diff
@@ -0,0 +1,648 @@
...
+            Ok((ct, shared_message))
+        }
+
+        /// Encrypt a shared message, adding the new encryption statement to the proof.
+        ///
+        /// This method purposefully reveals that two ciphertexts enrypt the same underlying value. If
+        /// this is not what you want, use [`Self::encrypt`].
+        ///
+        /// This method assumes that you've created the `message` argument with _this_ builder.
+        pub fn encrypt_shared(
```
ryanorendorff, line 417: `encrypt_and_share` and `encrypt_shared` are similar names, maybe we could have something slightly different?

sunscreen_runtime/src/linked.rs
```diff
@@ -343,10 +294,216 @@ impl<Q: Ring + CryptoHash + ModSwitch<ZqRistretto> + RingModulus<4> + Ord> Linke
...
+impl_from!(ZqSeal128_4096, LP3);
+impl_from!(ZqSeal128_8192, LP4);
+
+macro_rules! seq_zq {
+    ($block:tt) => (
+        seq!(N in 1..=4 {
+            $block
+        })
+    )
+}
```
ryanorendorff, line 389: What is this doing?

ryanorendorff (approved):

logproof/src/linear_relation.rs
ryanorendorff: Ya makes sense.

logproof/src/linear_relation.rs
ryanorendorff: It should never be 1, correct. 0 is an odd case in this function to essentially remove that element from the mixed bounds check.

logproof/src/linear_relation.rs
```diff
@@ -1120,6 +1120,7 @@ impl LogProof {
      * This modifies bitvec in place.
      *
      */
+    // TODO seems incorrect for log_b == 1; {0, 1} -> 0 and {-1} -> 1
```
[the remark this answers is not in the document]
samtay: Resolved in favor of conversation here https://github.com/Sunscreen-tech/Sunscreen/pull/339#discussion_r1475188504

Will probably add an assertion before merging :+1:

sunscreen/src/compiler.rs
```diff
@@ -398,37 +458,131 @@ impl<T, B> GenericCompiler<T, B> {
...
+        FheZkpCompiler {
+            data: CompilerData::new_fhe_zkp(fhe_data, self.data.unwrap_zkp()),
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Add the given ZKP program for compilation.
+    ///
+    /// Note that this method will not accept "shared" ZKP programs, which have inputs shared from
+    /// FHE programs. You must first call `fhe_program` to add an FHE program to unlock this
```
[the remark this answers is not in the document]
samtay: `fhe_program` is always callable, so the note doesn't apply there. And yes the verbiage is correct, you need to have called `fhe_program`, and then you'll have a compiler whose `.zkp_program` accepts shared ZKPs. Otherwise you'll get a Rust-compile-time error.

sunscreen/src/compiler.rs
```diff
@@ -542,17 +646,35 @@ where
...
         F: ZkpProgramFn<B> + 'static,
     {
-        self.data
-            .zkp_data_mut()
-            .zkp_program_fns
-            .push(Box::new(zkp_program_fn));
+        let zkp_data = self.data.zkp_data_mut();
+        match zkp_program_fn.into_shared() {
+            Ok(shared) => zkp_data.shared_zkp_program_fns.push(shared),
+            Err(not_shared) => zkp_data.zkp_program_fns.push(not_shared),
```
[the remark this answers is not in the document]
samtay: It is a bit odd, but because the method is `into_shared`, the return type seems reasonable to me. I thought it would be weird to create an either type isomorphic to result just for this but I guess we could?

sunscreen/src/compiler.rs
```diff
@@ -398,37 +458,131 @@ impl<T, B> GenericCompiler<T, B> {
...
+            .chain(shared_zkp_programs)
             .collect::<Result<HashMap<_, _>>>()?;
 
         Ok(zkp_programs)
     }
+}
 
-    fn compile_internal(self) -> Result<Application<T>> {
-        Application::new(HashMap::new(), self.compile_zkp()?)
+impl<B> ZkpCompiler<B>
```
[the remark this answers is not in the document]
samtay: It might help to think about the return value of `Compiler::compile`. The idea is to return `Application<Fhe/Zkp>` based on what was passed to the `Compiler`. You start with a blank `Compiler::new()` that can't even `compile()` into anything. You can _always_ add an `fhe_program` to your compiler, and when you do we make sure you get back an `Application<M>` where `M: marker::Fhe`.

It confused me too at first, but then I spent a long time in this file :slightly_smiling_face:

sunscreen/src/types/zkp/bfv_plaintext.rs
samtay: I have no idea where this came from. Copilot? :smile:

sunscreen/src/types/zkp/program_node.rs
samtay: There's no way for the user to add them out of order if that's what you mean. I can add a comment to the one place where inputs get combined in `sunscreen_runtime::linked` to document the internal requirement.

sunscreen/tests/linked.rs
samtay: I have some that will come in a new PR with additional shared input types.

sunscreen_runtime/src/builder.rs
samtay: I have no idea, this was added to the existing builders at some point, so I left it as is.

sunscreen_runtime/src/builder.rs
samtay: Well, I suppose for this trait that isn't a _strict_ requirement in general. But currently all the integer encodings use coefficients multiplied by powers of 2. So attempting the decoding at `N = 256` would result in `2^N` overflowing. I can reword this to make it more clear.

sunscreen_runtime/src/linked.rs
samtay: This is making the manual dynamic dispatch wrappers more maintainable. I had to create these
```rust
pub(crate) enum SealSdlpProverKnowledgeInternal {
    LP1(LogProofProverKnowledge<ZqSeal128_1024>),
    LP2(LogProofProverKnowledge<ZqSeal128_2048>),
    LP3(LogProofProverKnowledge<ZqSeal128_4096>),
    LP4(LogProofProverKnowledge<ZqSeal128_8192>),
}

#[derive(Debug, Clone)]
pub(crate) enum SealSdlpVerifierKnowledgeInternal {
    LP1(LogProofVerifierKnowledge<ZqSeal128_1024>),
    LP2(LogProofVerifierKnowledge<ZqSeal128_2048>),
    LP3(LogProofVerifierKnowledge<ZqSeal128_4096>),
    LP4(LogProofVerifierKnowledge<ZqSeal128_8192>),
}
```
to go from dynamic parameter sets to compile time parameter sets. That's a lot of match statements and a lot of boilerplate to maintain if we support more/less parameter sets or if we have to expose more functions.

sunscreen_runtime/src/builder.rs
samtay: In the PR desc I talked about changing `share` to `link` but other than that, I struggle to think of better names. There's `encrypt_again` rather than `encrypt_shared` that seems even worse. `reencrypt`? :shrug:

sunscreen/src/compiler.rs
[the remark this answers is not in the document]
ryanorendorff: Eh up to you.

sunscreen_runtime/src/builder.rs
ryanorendorff: I had added the prove with parameters as a stop gap to manually pass the parameters around but if we are doing it automatically I don't think the user needs to see this function.

sunscreen_runtime/src/builder.rs
ryanorendorff: Oh I was thinking a 256 degree polynomial. Whoops.

sunscreen_runtime/src/builder.rs
samtay: Ok I've removed them from the builder and removed a runtime `prove_with_parameters` that was unused. Marked the `verify_with_parameters` as `pub(crate)`.

logproof/src/linear_relation.rs
samtay: Added an assertion

PR #341, opened 2024-02-04
Link unsigned and rational FHE types

samtay:


sunscreen/tests/linked.rs
```diff
@@ -162,6 +163,150 @@ mod linked_tests {
...
+    fn compare_rational<F: FieldSpec>(#[linked] x: BfvRational<F>, #[linked] y: BfvRational<F>) {
+        let (x_n, x_d) = x.into_field_elem();
+        let (y_n, y_d) = y.into_field_elem();
+        let x = x_n * y_d;
+        let y = y_n * x_d;
+        x.constrain_le_bounded(y, 64)
+    }
+
+    #[test]
+    #[ignore = "bug in linked input multplication"]
```
samtay, line 268: This issue was discovered in these tests, but it is preexisting and independent of the work in this PR. Will be tackled next week.

sunscreen_runtime/src/linked.rs
```diff
@@ -89,6 +89,7 @@ fn new_single_party_with_shared_generators(
         if left_side {
             g[index] = *gen;
             left_side = false;
+            // TODO this overflows if shared arguments are provided with no constraints
```
samtay, line 92: Similar to above, something that needs to be tackled when we shore up the shared commitments, but unrelated to this PR.

sunscreen/tests/linked.rs
ryanorendorff: Ah I see! I bet the multiplication is throwing off the assumed generator order by adding a new input prior to where the link component thinks the multipliers will be.

sunscreen_runtime/src/linked.rs
ryanorendorff: As in the ZKP takes in a shared input but does nothing with them?

ryanorendorff (approved):
lgmt!

sunscreen_runtime/src/linked.rs
samtay: Yea exactly

PR #345, opened 2024-02-11
Disallow repeating linked inputs

samtay:
I didn't see a way to make this possible in logproof, so I'm disallowing it from the API.


ryanorendorff (approved):
LGTM!

PR #346, opened 2024-02-12
Diverge link methods

samtay:
I had merged the notions of referencing an existing message in SDLP and in ZKP. This doesn't make sense and also imposes unnecessarily limitations. This PR separates the two notions.

# api
Essentially the API looks like this (with each one also having a corresponding symmetric version):

```rust
impl Builder {
    /// Encrypt a plaintext, adding the encryption statement to the proof.
    pub fn encrypt<P>(&mut self, message: &P, public_key: &'k PublicKey) -> Result<Ciphertext>
        where
            P: TryIntoPlaintext + TypeName;

    /// Encrypt a plaintext, adding the encryption statement to the proof and returning the
    /// message to optionally be [encrypted again](`Self::encrypt_again`), that is, _shared_
    /// with another proof statement.
    pub fn encrypt_and_share<P>(&mut self, message: &P, public_key: &'k PublicKey) -> Result<(Ciphertext, Message)>
        where
            P: TryIntoPlaintext + TypeName;

    /// Encrypt a plaintext intended for linking.
    ///
    /// The returned `LinkedMessage` can be used:
    /// 1. to add an encryption statement of ciphertext equality to the proof (see [`Self::encrypt_again`]).
    /// 2. as a linked input to a ZKP program (see [`Self::linked_input`]).
    pub fn encrypt_and_link<P>(&mut self, message: &P, public_key: &'k PublicKey) -> Result<(Ciphertext, LinkedMessage)>
        where
            P: LinkWithZkp + TryIntoPlaintext + TypeName;

    /// Encrypt an existing message, adding the new encryption statement to the proof.
    ///
    /// This method purposefully reveals that two ciphertexts encrypt the same underlying value. If
    /// this is not what you want, use [`Self::encrypt`].
    pub fn encrypt_again<E>(&mut self, message: &E, public_key: &'k PublicKey) -> Result<Ciphertext>
        where
            E: ExistingMessage;
}
```

# other ideas
I played with some other ideas, like using a singular `Message<T = ()>` and `type LinkedMessage = Message<Link>` rather than the sealed `ExistingMessage` trait. We do use that pattern elsewhere so it wouldn't be inconsistent to do so, but in the end I thought above was simpler. I also considered always returning a `Message` from these methods, where `Message` can be converted into a ciphertext, but again thought it was not as clear as the above.

Let me know what you think regarding `ExistingMessage` vs `Message<T>`, I could go either way.

### DIFF
--- a/sunscreen_runtime/src/builder.rs
+++ b/sunscreen_runtime/src/builder.rs
@@ -253,47 +253,87 @@ mod linked {
         const DEGREE_BOUND: usize;
     }
 
-    #[derive(Debug, Clone)]
-    /// We pass this around for both plain and linked messages, as we need both the plaintext and
-    /// its type information to perform encryption.
+    /// We pass this around for both plain and linked messages. This type exists because we need
+    /// both the plaintext and its type information to perform encryption.
     ///
     /// Essentially we store the output from TryIntoPlaintext and TypeName.
-    pub(crate) struct PlaintextTyped {
+    #[derive(Debug, Clone)]
+    struct PlaintextTyped {
         plaintext: Plaintext,
         type_name: Type,
     }
 
-    /// We pass this around for just linked messages, as also need the zkp type information.
-    ///
-    /// Essentially we store the output from LinkWithZkp
-    #[derive(Debug, Clone)]
-    pub(crate) struct LinkedPlaintextTyped {
-        plaintext_typed: PlaintextTyped,
-        zkp_type: Type,
+    /// This is the internal type used for existing messages which have been added to the SDLP already.
+    /// The two public types [`Message`] and [`LinkedMessage`] are wrappers around this one.
+    #[derive(Clone, Debug)]
+    struct MessageInternal<Z = ()> {
+        id: usize,
+        len: usize,
+        pt: Arc<PlaintextTyped>,
+        zkp_type: Z,
     }
 
-    /// A [`Plaintext`] message that can be linked. Create this with [`LogProofBuilder::encrypt_and_link`].
+    /// A [`Plaintext`] message that can be [encrypted again](`LogProofBuilder::encrypt_again`).
+    #[derive(Clone, Debug)]
+    pub struct Message(MessageInternal<()>);
+
+    /// A [`Plaintext`] message that can be [encrypted again](`LogProofBuilder::encrypt_again`) or
+    /// [linked to a ZKP](`LogProofBuilder::linked_input`). Create this with
+    /// [`LogProofBuilder::encrypt_and_link`].
     #[derive(Debug)]
-    pub struct LinkedMessage {
-        pub(crate) id: usize,
-        pub(crate) message: Arc<LinkedPlaintextTyped>,
-        pub(crate) len: usize,
-    }
+    pub struct LinkedMessage(MessageInternal<Type>);
 
     impl LinkedMessage {
-        // Only allow cloning internally.
-        fn clone(&self) -> Self {
-            LinkedMessage {
-                id: self.id,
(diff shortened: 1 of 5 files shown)


sunscreen_runtime/src/builder.rs
```diff
@@ -398,119 +445,124 @@ mod linked {
             P: TryIntoPlaintext + TypeName,
         {
             let pt = self.plaintext_typed(message)?;
-            self.encrypt_symmetric_internal(Message::Plain(pt), private_key, None)
+            self.encrypt_symmetric_internal(Msg::from(pt), private_key, None)
         }
 
-        /// Encrypt a plaintext intended for sharing.
+        /// Encrypt a plaintext, adding the encryption statement to the proof and returning the
+        /// message to optionally be [encrypted again](`Self::encrypt_again`), that is, _shared_
```
ryanorendorff, line 452: I am not sure how to update this comment but I think that "adding the encryption statement to the proof" is tripping me up, because I don't know whether that is the SDLP proof or the bullet proof (I know now after we chatted about it). It would reveal low level details to say "add the encryption statement to the SDLP proof", so I don't like that, but without more details on which proof we are talking about it is a tad confusing.

ryanorendorff (approved):
This is mostly bike shedding around which proof system is being referenced. Potentially we could call `encrypt_and_share` as `encrypt_and_reuse_message` and `encrypt_and_link` could be `encrypt_and_send_to_zkp_program`. I don't know if there is an awesome way to differentiate the two ZKP methods without making it confusing to the user that there are two, but maybe that is because we know too much 😆 .

sunscreen_runtime/src/builder.rs
samtay: I don't think it's too low level to say SDLP proof, elsewhere in the API docs we talk about the linked proof linking the SDLP and the ZKP together.

samtay: Updated the comments

samtay:
Updated the naming to the following:

```rust
impl Builder {
    /// Encrypt a plaintext, adding the encryption statement to the logproof.
    pub fn encrypt<P>(&mut self, message: &P, public_key: &'k PublicKey) -> Result<Ciphertext>
        where
            P: TryIntoPlaintext + TypeName;

    /// Encrypt a plaintext, adding the encryption statement to the logproof and returning the
    /// message to optionally be [encrypted again](`Self::encrypt_again`), that is, _shared_
    /// with another proof statement.
    pub fn encrypt_returning_msg<P>(&mut self, message: &P, public_key: &'k PublicKey) -> Result<(Ciphertext, Message)>
        where
            P: TryIntoPlaintext + TypeName;

    /// Encrypt a plaintext intended for linking.
    ///
    /// The returned `LinkedMessage` can be used:
    /// 1. to add an encryption statement of ciphertext equality to the logproof (see [`Self::encrypt_again`]).
    /// 2. as a linked input to a ZKP program (see [`Self::linked_input`]).
    pub fn encrypt_returning_link<P>(&mut self, message: &P, public_key: &'k PublicKey) -> Result<(Ciphertext, LinkedMessage)>
        where
            P: LinkWithZkp + TryIntoPlaintext + TypeName;

    /// Encrypt an existing message, adding the new encryption statement to the logproof.
    ///
    /// This method purposefully reveals that two ciphertexts encrypt the same underlying value. If
    /// this is not what you want, use [`Self::encrypt`].
    pub fn encrypt_msg<E>(&mut self, message: &E, public_key: &'k PublicKey) -> Result<Ciphertext>
        where
            E: ExistingMessage;
}
```

PR #348, opened 2024-02-15
Decrypt statements

samtay:
Support "decryption statements", i.e. statements about ciphertexts that aren't fresh encryptions. These have the same format as symmetric statements in logproof, but the bounds are much more lax as errors can go up to $\Delta/2$ rather than being a fresh noise polynomial.


examples/private_tx_linkedproof/src/main.rs
```diff
@@ -1,69 +1,510 @@
...
+/// In this simple example, assume read-only references `&Chain` provide "call" functionalities,
+/// i.e. non-mutating methods for reading chain data, and mutable references `&mut Chain` provide
+/// "send" functionalities, i.e. sending transactions that can mutate chain data.
+pub struct Chain {
+    /// The current balances
+    balances: HashMap<Username, Ciphertext>,
+    /// The user's public keys
+    keys: HashMap<Username, PublicKey>,
+    /// Ledger of transactions
+    // TODO log transactions
```
ryanorendorff, line 267: Is this still a valid todo?

logproof/src/bfv_statement.rs
```diff
@@ -382,11 +440,16 @@ where
...
-    let m_default_bound = Bounds(vec![params.plain_modulus(); degree]);
+    let m_default_bound = Bounds(vec![params.plain_modulus().ceil_log2(); degree]);
     let r_bound = m_default_bound.clone();
     let u_bound = Bounds(vec![U_COEFFICIENT_BOUND; degree]);
     let e_bound = Bounds(vec![E_COEFFICIENT_BOUND; degree]);
     let s_bound = Bounds(vec![S_COEFFICIENT_BOUND; degree]);
+    // very liberal bound, the max that satisfies correctness
+    let q_div_2_bits = calculate_ciphertext_modulus(params.ciphertext_modulus())
+        .div(NonZero::from_uint(Uint::from(2u8)))
+        .ceil_log2();
```
ryanorendorff, line 451: Shouldn't this be delta/2, not q/2?

logproof/src/math.rs
```diff
@@ -92,14 +92,20 @@ pub trait InfinityNorm {
...
 {
     type Output = R;
 
     fn infinity_norm(&self) -> Self::Output {
-        self.coeffs
-            .iter()
-            .fold(R::zero(), |max, x| if x > &max { x.clone() } else { max })
+        self.coeffs.iter().fold(R::zero(), |max, x| {
+            let x_neg = x.clone().neg();
+            let x_abs = x.min(&x_neg);
```
ryanorendorff, line 102: Should this be max?

ryanorendorff (approved):

logproof/src/math.rs
samtay: I guess it depends on your intent, and more technically it depends on the ring's `Ord` impl. Because of our 2s complement encoding in logproof, this number is more relevant to us. (It matches what we've [done here](https://github.com/Sunscreen-tech/Sunscreen/blob/4db46f0967c8862c55928cf46a8a14ce4a76861b/logproof/src/linear_relation.rs#L1145) anyways).

samtay: Sorry, I mispoke, I'm nearing brain dead over here. The above is generally correct for all of our rings which defer their ord instance to their underlying unsigned integer representation. Valid values are in the range `[0..R::MODULUS]` with `R::MODULUS/2` being the negative cutoff. So the absolute value of `x` will always be the minimum of `x, x.neg()` as it one will be in the first half of the range, and the other in the greater half of the range.

logproof/src/bfv_statement.rs
samtay: Thanks! Fixed.

PR #362, opened 2024-02-28
Compiler params

samtay:
Allow setting FHE params without FHE programs.

This changes the semantic meaning of the markers on the `Compiler` struct. Now `Compiler<T: Fhe>` means that the compiler has derivable FHE parameters, either through manually set params or through FHE program arguments.

I've also moved the program getters onto the generic `Application` type... calling `app.get_fhe_programs()` and getting an empty hashmap if you haven't provided FHE programs is quite reasonable behavior.

The only type safety lost really is that now you _could_ call `Compiler::with_params().compile()` and get a runtime `Err(NoPrograms)` rather than a compile-time error. But imo providing type safety against such an error is not very valuable.

Depends on #358 / #361


sunscreen/src/compiler.rs
```diff
@@ -213,51 +205,30 @@ impl<B> CompilerData<B> {
...
-// But there's one added complication in that the ZKP method `.zkp_program` has type restrictions
-// depending on if the current ZKP-capable compiler is also FHE-capable, so that method also needs
-// multiple definitions.
-//
-// TODO: allow FHE params to be specified for both FHE and ZKP programs. This is
-// surprisingly complicated with the current approach. We may have to split into two markers (one
-// for FHE and one for ZKP) for that to be feasible, without requiring 6 definitions for the same
-// method.
+// - FHE: indicates params are derivable, either through FHE programs or manual params setting.
+// - ZKP: indicates a ZKP backend has been specified, thus user can add ZKP programs.
```
ryanorendorff, line 231: Should this comment be FHEZKP instead of ZKP? Since `CompilerData` specifies B which can be `Fhe` or `FheZkp`

ryanorendorff (approved):
lgtm

sunscreen/src/compiler.rs
samtay: Actually no, sorry this is maybe a tad more confusing now. This comment is referring to the meanings of the marker traits, not the shape of the compiler data.

samtay: Maybe it would be a little cleaner if we kept the `CompilerData::{None, Zkp}` variants, but pulled out the `FheCompilerData { params_mode }` field so that it is always present. Then the shape of the compiler data would match the marker more closely.

PR #368, opened 2024-03-04
Rweber/tfhe mem

rickwebiii:
Simplify scratch allocation, used `AVec` to meet runtime alignment requirements

### DIFF
--- a/sunscreen_tfhe/src/entities/glwe_ciphertext.rs
+++ b/sunscreen_tfhe/src/entities/glwe_ciphertext.rs
@@ -74,7 +74,7 @@ where
         assert_eq!(data.len(), GlweCiphertextRef::<S>::size(params.dim));
 
         GlweCiphertext {
-            data: AVec::from_iter(SIMD_ALIGN, data.iter().map(|x| Torus::from(*x))),
+            data: avec_from_iter!(data.iter().map(|x| Torus::from(*x))),
         }
     }
 }
--- a/sunscreen_tfhe/src/entities/glwe_ciphertext.rs
+++ b/sunscreen_tfhe/src/entities/glwe_ciphertext.rs
@@ -1,12 +1,9 @@
+use aligned_vec::AVec;
 use num::{Complex, Zero};
 use serde::{Deserialize, Serialize};
 
 use crate::{
-    dst::{FromMutSlice, FromSlice, OverlaySize},
-    entities::GgswCiphertextRef,
-    macros::{impl_binary_op, impl_unary_op},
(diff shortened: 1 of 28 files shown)


sunscreen_tfhe/src/scratch.rs
```diff
@@ -91,32 +100,25 @@ where
...
-impl Drop for Scratch {
-    fn drop(&mut self) {
-        let top = unsafe { Box::from_raw(self.top) };
-
-        std::mem::drop(top);
-    }
+    stack: *mut LinkedList<*mut Allocation>,
 }
 
 impl Scratch {
```
rickwebiii, line 113: We need a `Drop` impl

sunscreen_tfhe/src/entities/glwe_ciphertext.rs
```diff
@@ -72,10 +74,7 @@ where
         assert_eq!(data.len(), GlweCiphertextRef::<S>::size(params.dim));
 
         GlweCiphertext {
-            data: data
-                .iter()
-                .map(|x| Torus::from(*x))
-                .collect::<Vec<Torus<S>>>(),
+            data: AVec::from_iter(SIMD_ALIGN, data.iter().map(|x| Torus::from(*x))),
```
ryanorendorff, line 77: Maybe this could also be made a macro or function so that the user can use it more like a standard `from_iter` method. Which is to say `SIMD_ALIGN` would be hidden.

sunscreen_tfhe/src/scratch.rs
```diff
@@ -126,102 +135,95 @@ impl Scratch {
...
-                let data = vec![0u8; req_len];
+                let allocation = Box::new(allocation);
+                Box::into_raw(allocation)
+            } else if (*allocation.unwrap()).data.alignment() < alignment
+                || (*allocation.unwrap()).data.len() < u8_len
+            {
+                // If we found an allocation, but its size and len requirements
+                // are insufficient.
+                let allocation = allocation.unwrap();
+                let drop_box = Box::from_raw(allocation);
```
ryanorendorff, line 158: I am guessing it doesn't make sense to look further down for an allocation that might have the right amount of space? This will just reallocate the first free thing it finds from what I am reading.

ryanorendorff (approved):
lgtm

sunscreen_tfhe/src/scratch.rs
rickwebiii: Correct. This is less complex and also has the MRU property, which should be slightly more cache efficient.

PR #369, opened 2024-03-06
Add cmux_tree tfhe test

ryanorendorff:
Test for an arbitrary cmux tree to see if the noise grows too much to decode.


ryanorendorff:
Closing since the test takes too long. However, the test does pass in CI.

PR #382, opened 2024-11-20
Extract out GLev encryption into its own operations

ryanorendorff:
Extracts out GLev encryptions into their own operations, and then uses these operations when performing GGSW operations.

This PR also changes the GGSW encryption of a scalar to convert it to a polynomial message first and then perform the encryption. This will slightly increase the encryption time for a GGSW ciphertext from a plaintext message, but reduces sources of bugs by only having one encryption method for GGSW.

### DIFF
--- a/sunscreen_tfhe/src/entities/glev_ciphertext.rs
+++ b/sunscreen_tfhe/src/entities/glev_ciphertext.rs
@@ -1,7 +1,9 @@
-use num::Complex;
+use num::{Complex, Zero};
 use serde::{Deserialize, Serialize};
 
-use crate::{dst::OverlaySize, GlweDef, GlweDimension, RadixCount, Torus, TorusOps};
+use crate::{
+    dst::OverlaySize, GlweDef, GlweDimension, RadixCount, RadixDecomposition, Torus, TorusOps,
+};
 
 use super::{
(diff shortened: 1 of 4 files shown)


sunscreen_tfhe/src/entities/glev_ciphertext.rs
```diff
@@ -29,6 +31,20 @@ where
     }
 }
 
+impl<S> GlevCiphertext<S>
+where
+    S: TorusOps,
+{
+    /// Create a new zero GGSW ciphertext with the given parameters.
```
rickwebiii, line 38: GGSW -> GLEV

sunscreen_tfhe/src/ops/encryption/glev_encryption.rs
```diff
@@ -0,0 +1,271 @@
...
+};
+
+use super::{
+    decrypt_glwe_ciphertext, encrypt_glwe_ciphertext_secret,
+    trivially_encrypt_glwe_with_sk_argument,
+};
+
+/// Perform a GLev encryption. This is generic in case a trivial GLev encryption
+/// is wanted (for example, for testing purposes).
+pub(crate) fn encrypt_glev_ciphertext_generic<S>(
```
rickwebiii, line 16: I think it's okay to export these publicly. Maybe some cryptographers wanna do some weird stuff with GLEVs?

rickwebiii (approved):

PR #384, opened 2024-12-05
Implement scheme switching

ryanorendorff:
Implements scheme switching, which allows GLevs to be converted to GGSW

# Background
Let $\mathcal{R}=\mathbb{Z_q}[X]/(X^N+1)$ for power of two $N$.

Recall that $\mathsf{GLEV}_{\vec{s}}(m)=[ \mathsf{GLWE}(\frac{q}{\beta^1}m), \mathsf{GLWE}(\frac{q}{\beta^2}m), ..., \mathsf{GLWE}(\frac{q}{\beta^\ell}m) ]$ where $\beta$ and $\ell$ are scheme parameters that define a radix decomposition.

Furthermore, recall the gadget product $\odot$ between $a \in \mathcal{R}$ and $\mathsf{GLEV}(m)$:

$$
a \odot \mathsf{GLEV}(m):=\sum_{i=0}^{\ell-1}\mathsf{Decomp}_{i, \beta}(a)\times\mathsf{GLWE}(\frac{q}{\beta^{i+1}}m)
$$
$$
\approx\mathsf{GLWE}(am)
$$

# Scheme switching
## Keygen
Given a GLWE scheme with poly degree $N$ and GLWE size $k$ and secret key $\vec{s}$, define a scheme switching key as follows:

* Let $\mathbf{sk} = \vec{s} \otimes \vec{s}$
* Compute scheme switching key $\mathbf{s_{ss}}$ where $\mathbf{s_{ss}}^{i,j}=\mathsf{GLEV_{\vec{s}}}(\mathbf{sk}_{i,j})$ for $i, j\in [0, k)$.
* Observe that since $\mathbf{sk_{i,j}}=\mathbf{sk_{j, i}}$, we can reduce our keysize by roughly half. Simply encrypt only the lower half of the matrix and mirror requests in the upper half. E.g. when an algorithm requests $i=2, j=3$, return the encryption for $i=3, j=2$

### First, an observation
Suppose we have $(\vec{a}, b) = \mathsf{GLWE}(m)$. Construct trivial GLWE ciphertext $t$ by placing $b$ in the $p$'th place in the basis coefficients and 0 elsewhere $t_p(b)=((0, ..., b, ... 0), 0)$. Observe what happens if we decrypt $t$ under any key $\vec{s}$:

$$
m = -(\sum_{i \ne p}^{[0, k)}0\cdot s_i + b \cdot s_p) - 0
$$

$$
= -b \cdot s_p
$$

Since the error is 0 as well, we can elide the rounding step. Thus, $t$ is a $\mathsf{GLWE}$ encryption of $m \cdot s_p$ under $\vec{s}$.

This result is from the convention that the encryption equation is $b = a \cdot s + m + e$, which is what we use in our TFHE variant. The "Circuit Bootstrapping: Faster and Smaller" paper uses the convention $b = -a \cdot s + m + e$; either is valid as long as the definition is consistent. The next section assumes the all positive convention we use.

### Our regularly scheduled program
Given $x=\mathsf{GLEV}(m)$, we have $x_i=\mathsf{GLWE}(\frac{q}{\beta^{i+1}}m)=(\vec{a}^{(i)}, b^{(i)}), i\in[0,\ell_{ggsw})$.

For each $i \in [0, \ell_{ggsw}), j \in [0, k)$ compute using $\mathsf{s_{ss}}^{j,r}$ where $r \in [0, k)$

$$
y_{i, j}=t_j(b^{(i)}) + \sum_{r=0}^{k-1} a^{(i)}_r \odot \mathsf{GLEV}_{\vec{s}}(s_j \cdot s_r)=\mathsf{GLWE}_{\vec{s}}(\sum_{r=0}^{k-1}a^{(i)}_r \cdot s_r \cdot s_j - b^{(i)}\cdot s_j)
$$
$$
=\mathsf{GLWE}_{\vec{s}}((\sum_{r=0}^{k-1}a^{(i)}_r \cdot s_m - b^{(i)})\cdot s_j)
$$
$$
=\mathsf{GLWE}_{\vec{s}}(-\frac{q}{\beta^{i+1}}\cdot m \cdot s_j + e_i \cdot s_j)
$$

Note the $e_i$ term is small if $s_j$ is small (i.e. binary), and thus we are left with encryptions of $-s_j \cdot m$

Further note, the radix decomposition in the above $\odot$ is $\beta_{ss}, \ell_{ss}$, which may be distinct from $(\beta_{ggsw}, \ell_{ggsw})$

Let $z_j=\mathsf{GLEV}_{\vec{s}}(m \cdot s_j)=(y_{0,j}, y_{1,j}, ..., y_{\ell_{ggsw}-1, j})$

Output $\mathsf{GGSW}_{\vec{s}}(m)=(z_0, z_1, ..., z_{k-1}, x)$

# References

- Wang, R., Wen, Y., Li, Z., Lu, X., Wei, B., Liu, K., & Wang, K. (2024, May). Circuit bootstrapping: faster and smaller. In Annual International Conference on the Theory and Applications of Cryptographic Techniques (pp. 342-372). Cham: Springer Nature Switzerland. [Circuit Bootstrapping Faster and Smaller.pdf](https://github.com/user-attachments/files/17986823/Circuit.Bootstrapping.Faster.and.Smaller.pdf)

- De Micheli, G., Kim, D., Micciancio, D., Suhl, A.: Faster amortized FHEW bootstrapping using ring automorphisms. Cryptology ePrint Archive (2023) [Faster Amortized FHEW Bootstrapping Using Ring Automorphisms.pdf](https://github.com/user-attachments/files/17986824/Faster.Amortized.FHEW.Bootstrapping.Using.Ring.Automorphisms.pdf)

### DIFF
--- a/sunscreen_tfhe/src/ops/bootstrapping/scheme_switch.rs
+++ b/sunscreen_tfhe/src/ops/bootstrapping/scheme_switch.rs
@@ -299,16 +299,6 @@ pub fn scheme_switch<S>(
             // where the b_i is at the j-th position.
             update_encrypted_secret_key_component(encoded_b_i, x_i, j, params);
 
-            // If we are in debug mode, we can assert that the ciphertext is
-            // correctly shaped.
-            #[cfg(debug_assertions)]
-            assert_encrypted_secret_key_component_correct_shape(
-                encoded_b_i,
-                x_i.b(params),
-                j,
-                params,
-            );
-
             // We can clone directly into y_i_j since this will overwrite it.
             y_i_j.clone_from_ref(encoded_b_i);
 
@@ -327,51 +317,6 @@ pub fn scheme_switch<S>(
             }
         }
     }
-}
-
-#[allow(unused)]
-pub(crate) fn assert_encrypted_secret_key_component_correct_shape<S>(
-    glwe: &GlweCiphertextRef<S>,
-    b: &PolynomialRef<Torus<S>>,
-    index: usize,
-    params: &GlweDef,
-) where
-    S: TorusOps,
-{
-    // Verify the output:
-    // 1. The mask (a) should be zero everywhere except at index
-    // 2. At index, it should contain the body of the input ciphertext
-    // 3. The body (b) should be zero
-
-    for (i, a_i) in glwe.a(params).enumerate() {
-        if i == index {
-            // At index, coefficients should match input body
-            assert_eq!(
-                a_i.coeffs(),
-                b.coeffs(),
-                "Coefficients at index {} don't match input body",
-                index
-            );
-        } else {
-            // All other positions should be zero
-            assert!(
-                a_i.coeffs()
-                    .iter()
-                    .all(|x| *x == Torus::from(<S as num::Zero>::zero())),
-                "Non-zero coefficients found at index {} (should be all zero)",
-                i
-            );
-        }
-    }
-
-    // Verify body is zero
-    assert!(
-        glwe.b(params)
-            .coeffs()
-            .iter()
-            .all(|x| *x == Torus::from(<S as num::Zero>::zero())),
-        "Output body contains non-zero values"
-    );
 }
 
 #[cfg(test)]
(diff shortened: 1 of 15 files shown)


sunscreen_tfhe/src/ops/bootstrapping/scheme_switch.rs
```diff
@@ -0,0 +1,834 @@
...
+                    ssk.get_glev_at_index(j, r, params, radix_ss),
+                    params,
+                );
+            }
+        }
+    }
+}
+
+#[allow(unused)]
+pub(crate) fn assert_encrypted_secret_key_component_correct_shape<S>(
```
rickwebiii, line 333: I would move this into the test module and make it private.

rickwebiii (approved):

PR #387, opened 2024-12-29
Rweber/publicrlwe

rickwebiii:
Add RLWE encryption using double-lwe encryption trick.

### DIFF
--- a/sunscreen_tfhe/src/ops/encryption/rlwe_encryption.rs
+++ b/sunscreen_tfhe/src/ops/encryption/rlwe_encryption.rs
@@ -1,8 +1,6 @@
 use crate::{
     dst::FromMutSlice,
-    entities::{
-        GlweCiphertext, GlweSecretKeyRef, Polynomial, PolynomialRef, RlwePublicKeyRef,
-    },
+    entities::{GlweCiphertext, GlweSecretKeyRef, Polynomial, PolynomialRef, RlwePublicKeyRef},
     ops::encryption::encrypt_glwe_ciphertext_secret,
     polynomial::{polynomial_add_assign, polynomial_external_mad},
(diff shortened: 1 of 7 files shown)


sunscreen_tfhe/src/params.rs
```diff
@@ -223,6 +223,15 @@ pub const LWE_512_128: LweDef = LweDef {
...
 
+/// 128-bit secure parameters for a GLWE instance with a dimension of 512.
+pub const GLWE_1_512_128: GlweDef = GlweDef {
+    dim: GlweDimension {
+        size: GlweSize(1),
+        polynomial_degree: PolynomialDegree(512),
+    },
+    std: Stddev(0.0004899836456140595),
+};
+
```
ryanorendorff, line 234: Add a test for this in `check_security_levels` so that if we end up changing the security level models we will know which parameters need changing.

sunscreen_tfhe/src/ops/encryption/rlwe_encryption.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    rlwe_encrypt_public(ct, encoded, public_key, glwe)
+}
+
+/// Encrypts an encoded polynomial message using the given [`RlwePublicKey`](RlwePublicKeyRef).
+///
+/// # Algorithm
+/// This method uses the double-lwe trick, which is efficient but introduces
+/// more noise than Regev's trick (i.e. summing many encryptions of zero).
```
ryanorendorff, line 89: If you have a reference for this in the literature it would be helpful to add. Or add the equation for how decrypting `(p0 * u + e0, m + p1 * u + e1)` works.

ryanorendorff (approved):

